PR #4, opened 2023-11-08
Feature: values from new datastore

mariaozamiz:
### :pushpin: References

-   **Issue:** Closes [Make d2-autogen-forms read values from the new datastore](https://app.clickup.com/t/862kham84)

### :memo: Implementation

A branch was created with the base of feat/mal-nhwa-customforms and then development was merged to incorporate all the new code.

A dataSet code was supposed to be added as the second parameter to the build method, but this method was also being used in other files (such as DHIS2DataElement) where dataSet code couldn't be retrieved (only dataSet ID was available). 

The second parameter was set as optional to work this around, as it's likely not necessary to pass the dataSet code to every element as the config is cached.

If anyone has alternative solutions or insights, please let me know to improve this PR.

### :video_camera: Screenshots/Screen capture

Captura de pantalla 2023-11-07 a las 21 01 52


MiquelAdell (commented):
Reviewed the pull request functionally. Will like a code review by @eperedo 

Can approve the PR afterwards.

Thanks all!

src/data/common/Dhis2DataStoreDataForm.ts
```diff
@@ -217,11 +217,14 @@ export class Dhis2DataStoreDataForm {
         this.subNationals = config.subNationals;
     }
 
-    static async build(api: D2Api): Promise<Dhis2DataStoreDataForm> {
+    static async build(api: D2Api, dataSetCode?: string): Promise<Dhis2DataStoreDataForm> {
         if (cachedStore) return cachedStore;
 
-        const dataStore = api.dataStore(dataStoreNamespace);
-        const storeValue = await dataStore.get<object>(Namespaces.AUTOGENERATED_FORMS).getData();
+        let dataStore = api.dataStore(dataStoreNamespace);
```
eperedo, line 223: Normally we try to avoid in place updates. We can use `const` here and then validate if `dataSetCode` does not exist.

```ts
if (!dataSetCode) throw Error(`Unable to load configuration: dataSetCode not defined`);
```

After that passing dataSetCode to load the config:

```ts
const storeValue = await dataStore.get<object>(dataSetCode).getData();
```

What do you think?

src/data/common/clients/storage/Namespaces.ts
```diff
@@ -1,4 +1,5 @@
-export const dataStoreNamespace = "d2-reports";
+export const dataStoreNamespace = "d2-autogen-forms";
+export const formerDataStoreNamespace = "d2-autogen-forms";
```
eperedo, line 2: For the new namespace we can use the `D2_AUTOGEN_FORMS` attribute under the `Namespaces` object.

```
// Line 7
export const Namespaces = {
    ...
    D2_AUTOGEN_FORMS: "d2-autogen-forms",
};
```

eperedo (changes requested):
Thanks @mariaozamiz. Code looks good just minor adjustments to avoid duplicate variables/conditions. Let me know if you have any doubts

eperedo (approved):

PR #5, opened 2023-11-13
Feat/grid with cat option combo

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/865d7afyn

### :memo: Implementation
- Added the categoryOptionCombos field to each data element
- Created a new viewType component that forms a table with the categoryOptionCombos as columns and dataElements as rows

### :video_camera: Screenshots/Screen capture

https://github.com/EyeSeeTea/d2-autogen-forms/assets/37223065/6e79ce11-f73d-4af7-b04f-12dbb1b5cefe

`REACT_APP_DHIS2_BASE_URL=https://dev.eyeseetea.com/who-dev-238/`
`REACT_APP_REPORT_VARIANT=admin`


MiquelAdell (commented):
It seems to work on my side.

I have used the instructions @deeonwuli provided and then visited this URL:
http://localhost:8081/?dataSetId=PWCUb3Se1Ie&dataSet=PWCUb3Se1Ie&orgUnit=hmZE3mVAZFf&period=2022

@eperedo could you provide a code review for that?

Thanks!

src/webapp/reports/autogenerated-forms/GridWithCatOptionCombos.tsx
```diff
@@ -0,0 +1,128 @@
...
+                                    )?.dataElement;
+
+                                    if (dataElement) {
+                                        return (
+                                            <DataTableCell key={[dataElement.id, column.name].join("-")}>
+                                                <DataElementItem
+                                                    dataElement={dataElement}
+                                                    dataFormInfo={dataFormInfo}
+                                                    noComment={
+                                                        dataElement.name !== "Source type for HWF - (Inputs & Outputs)"
```
eperedo, line 104: This validation was for other viewType. Since we don't want to show the comments let's just pass `true` or leave the prop without any value.

```tsx
  <DataElementItem 
      dataElement={dataElement}
      dataFormInfo={dataFormInfo}
      // same as noComment={true}
      noComment
```

src/webapp/reports/autogenerated-forms/GridWithCatOptionCombosViewModel.ts
```diff
@@ -0,0 +1,93 @@
...
+const separator = " - ";
+
+export class GridWithCatOptionCombosViewModel {
+    static get(section: Section): Grid {
+        const subsections = _(section.dataElements)
+            .flatMap(dataElement => {
+                const cocNames = dataElement.categoryOptionCombos.map(coc => coc.name);
+                return cocNames.flatMap(coc => ({
+                    ...dataElement,
+                    cocId: dataElement.categoryCombos.categoryOptionCombos.find(c => c.name === coc)?.id || "cocId",
```
eperedo, line 41: I don't understand why are we returning `cocId` if we don't found a `categoryOptionCombos`? Since we are sending this ID to the dataValues endpoint we should make sure to always send the right id. As an alternative we can save undefined and then do a `compact` so we ignore the whole item.

eperedo (changes requested):
Thanks @deeonwuli code looks good, a couple of observations about a hardcode validation and an error I'm getting every time I tried to save a value.

image

See the `co` parameter is equal to `cocId` instead the category option combo id.

src/webapp/reports/autogenerated-forms/GridWithCatOptionCombos.tsx
```diff
@@ -0,0 +1,126 @@
...
+                                ) : (
+                                    <DataTableCell colSpan="2">
+                                        <span>{row.deName}</span>
+                                    </DataTableCell>
+                                )}
+
+                                {grid.columns.map(column => {
+                                    const dataElement = column.items.find(item =>
+                                        item.row.rows.map(row => row.deName).includes(row.deName)
+                                    )?.dataElement;
```
eperedo, line 95: It's a good practice to not include logic inside UI component. For this case I think we can move the code to the `GridWithCatOptionCombosViewModel.ts` file

Also here we're having the same dataElement for all the **Microscopy (All ages, both active and passive)** section. You can inspect the ids in devtools:

This is correct the dataElement id for **Microscopy (All ages, both active and passive) - Tested** is **DPNsg3wjxGm**
image

but if you look the other rows, all of them (except **Mixed (more than one species detected)**) are being assigned to the same data element.

image

For dataElement **Microscopy (All ages, both active and passive) - Positive** it should be the ID **enVf9I39FRZ**

eperedo (changes requested):
@deeonwuli thanks for the code, now it's generating the right category option combos,
but the dataElements are not right. Take a look to my comment in the `GridWithCatOptionCombos.tsx` file.

Also in the original custom form we have a row with the name "Microscopy   All ages, both active and passive", 

image

but in autogenforms we have that text as a column

image

If the requirement change that's ok just curious about the difference.

Let me know if you have any questions!

eperedo (approved):
Looks good! Thank you @deeonwuli!

PR #11, opened 2024-01-16
change column description based on de value from another section

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/8693fnxz5

### :art: Screenshots

https://github.com/EyeSeeTea/d2-autogen-forms/assets/37223065/c81c34fc-a07d-4ede-9d11-a4c505d5f032

Datastore config:
```
"columnsDescriptions": {
    "Combined": "<%= MAL_SECTOR_DATA_COMBINED %> Only data that cannot be disaggregated into separate health sectors",
    "Community": "Only data from the community health workers",
    ...
  }
```
where `Combined` and `Community` are the column names and `MAL_SECTOR_DATA_COMBINED` is the code of the data element used to change the value of the column description. 

`REACT_APP_DHIS2_BASE_URL=https://dev.eyeseetea.com/who-dev-238/
`


MiquelAdell:
I was not able to reproduce.

I set myself in the feat/change-column-description-text-with-de-value branch. 
Ran yarn start and visited http://localhost:8081/?dataSetId=PWCUb3Se1Ie&dataSet=PWCUb3Se1Ie&orgUnit=hmZE3mVAZFf&period=2022

I'm almost sure that I'm at the correct data set as I can see the changes in the datastore here:
image

But I can not see the column descriptions here:
Screenshot 2024-01-17 at 09 04 39

I was trying to make sure that this worked also by using constants. If it works as displayed in the video is ok but we'll also need the other so we can translate the strings.

Can you please confirm that it also works that way and see if there is something I'm doing wrong that explains why I can see the descriptions on my side?

Thanks!

MiquelAdell (approved):
It's ok on my side

I'm not sure if @eperedo already provided feedback because he suggested referencing other fields by code instead of id or something like that. @eperedo if the PR is ok by you as it is let me know and I'll merge it

Thanks!

eperedo (changes requested):
@deeonwuli thanks for the code, it works great.
Just a couple of minor things to avoid code duplication.

eperedo (approved):
looks good, thank you @deeonwuli

PR #20, opened 2024-02-06
import existing translations

eperedo:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/8693hjjkh

### :memo: Implementation

```bash
  npx ts-node src/scripts/translations-mal-wmr.ts -u username:password \
  --url http://localhost:8080 \
  --path=./translations.json \ # json file with all the translations (attached in the issue)
  --post true \ # persist changes into DHIS
  --export true \ # save metadata file to disk
  --autogenform ./autogenform.json # save autogenform config for dataElements
```


MiquelAdell (approved):
it does work, thanks!

PR #24, opened 2024-02-27
Feat/extend autogen configurator functionality

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/8693wpcbd

### :memo: Implementation
This PR extends the autogen-configurator to include 
- `columnsDescriptions`
- `matrixGrid` viewType

`REACT_APP_DHIS2_BASE_URL=https://dev.eyeseetea.com/who-dev-238/
`


src/data/autogenerated-forms-configurator/DataStoreConfigurationRepository.ts
```diff
@@ -46,11 +47,127 @@ export class Dhis2DataStoreConfigurationRepository implements DataStoreConfigura
...
+                },
+                paging: false,
+                rootJunction: "OR",
+            })
             .getData();
 
+        return dataSets;
+    }
+
+    async getSelectedDataSet(dataSets: DataSet[], dataSetCode: Maybe<Code>): Promise<AutogenConfigSchema> {
```
eperedo, line 68: Having a method `getSelectedDataSet` inside `DataStoreConfigurationRepository` could be confusing. Normally a repository should work with an specific Entity.

My proposal would be:

This whole method should be in the `GetSelectedDataSetUseCase` file instead here.

```ts
export class GetSelectedDataSetUseCase {
    constructor(private constantsRepository: ConstantRepository) {}

    execute(dataSets: DataSet[], dataSetCode: Maybe<Code>) Promise<AutogenConfigSchema> {
        const dataSet = _.find(dataSets, { code: dataSetCode });
        if (!dataSetCode || !dataSet)
            return { sections: [], categoryComboCodes: [], dataElements: [], constantCodes: [] };

        // we need a get method in the constant repository
        const constantsCodes = await this.constantRepository.get();
        const dataElements = this.getDataElements(dataSet);
        const sections = this.getSections(dataSet);
        const categoryComboCodes = this.getCategoryComboCodes(dataSet);

        return {
            categoryComboCodes: categoryComboCodes,
            constantCodes: constantCodes,
            dataElements: dataElements,
            sections: sections,
        };
    }
}
```

and maybe changing the filename could be `GetAutogenConfigSchemaUseCase.ts` since that's the object we're building.

Does it make sense?

src/data/autogenerated-forms-configurator/DataStoreConfigurationRepository.ts
```diff
@@ -46,11 +47,127 @@ export class Dhis2DataStoreConfigurationRepository implements DataStoreConfigura
...
-            .get<{ constants: { code: string }[] }>("/constants?fields=id,code")
+    private async getMetadata() {
+        const { objects: dataSets } = await this.api.models.dataSets
+            .get({
+                fields: dataSetFields,
+                filter: {
+                    name: {
+                        ilike: "NHWA",
+                        like: "MAL",
+                    },
```
eperedo, line 58: This filter means the app only works with these dataSets? (if that's a requirement is fine, just wondering since we're trying to do a more generic app)

src/domain/autogenerated-forms-configurator/entities/AutogenConfig.ts
```diff
@@ -0,0 +1,12 @@
+export interface AutogenConfig {
+    categoryCombinations: unknown;
+    dataElements: unknown;
+    dataSets: unknown;
+}
+
+export type Code = string;
+
```
eperedo, line 8: We already have a Code type in `entities/Base.tsBase.ts`, can we reuse it here?

eperedo (changes requested):
Thanks @deeonwuli. Some details about clean architecture. Let me know if you have any questions.

src/data/autogenerated-forms-configurator/DataStoreConfigurationRepository.ts
deeonwuli: yes. i used this filter as i assumed we would only be working with MAL and NHWA datasets. could @MiquelAdell confirm please?

MiquelAdell: @deeonwuli right now we are only working with these datasets but the idea is to be as generic is possible.
Why is this needed?

deeonwuli: it is not needed then. thank you @MiquelAdell

src/domain/common/repositories/ConstantRepository.ts
```diff
@@ -1,7 +1,9 @@
+import { Code } from "../entities/Base";
 import { Constant } from "../entities/Constant";
 import { SaveOptions } from "../entities/SaveOptions";
 import { Stats } from "../entities/Stats";
 
 export interface ConstantRepository {
+    get(): Promise<Code[]>;
```
eperedo, line 7: Normally a repository work with an Entity. In this case the `get` method should return an array of `Constant[]`.

src/data/ConstantD2Repository.ts
```diff
@@ -12,6 +12,12 @@ import { writeFileSync } from "fs";
 export class ConstantD2Repository implements ConstantRepository {
     constructor(private api: D2Api) {}
 
+    async get(): Promise<string[]> {
+        const { constants } = await this.api.metadata.get({ constants: { fields: { code: true } } }).getData();
+
+        return constants.filter(constant => constant.code).map(constant => constant.code);
```
eperedo, line 18: Right now this implementation works for the `GetAutogenConfigSchemaUseCase` usecase because in there we need all the codes. But if in the future we need an array of constants (not just the code or any specific attribute) in other usecase we'll need to implement a new method. 

So my proposal here is to return an array of Constants[] and then in the usecase add the logic to only get the codes. Making our implementation more generic and easy to use for any future usecase. Does it make sense?

eperedo (changes requested):
Thanks @deeonwuli we're very close. An observation in order to keep the ConstantRepository a bit more generic. Let me know if you have any questions.

eperedo (approved):
Thanks @deeonwuli!

MiquelAdell (approved):
tested and working, thanks!

PR #29, opened 2024-03-12
compile autogen configurator as app

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869413jx7

```bash
yarn build-configurator
```

This will create a `d2-autogen-configurator.zip` folder at the root of project that can be uploaded as an app in DHIS2. 
The `appVersions` property is used to configure the version of the app independent of the `package.json` version.


eperedo (changes requested):
Thanks @deeonwuli the build process works, but only if you have in your `.env` the value `REACT_APP_REPORT_VARIANT=autogenerated-forms-configurator`. 

can we include this value in the `build-configurator` script? 

```json
"build-configurator": "REACT_APP_REPORT_VARIANT=autogenerated-forms-configurator ..."
```

eperedo (approved):
looks good to me, thanks @deeonwuli!

PR #32, opened 2024-03-22
change grid-with-cat-option-combos column title catOptionCombo form name

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/86944kn4z

### :art: Screenshots
Screenshot 2024-03-22 at 01 13 12

```bash
REACT_APP_DHIS2_BASE_URL=https://dev.eyeseetea.com/who-dev-238/
```

### DIFF
--- a/src/data/common/Dhis2DataElement.ts
+++ b/src/data/common/Dhis2DataElement.ts
@@ -48,16 +48,20 @@ const dataElementFields = {
             name: true,
             categoryOptions: {
                 id: true,
+                code: true,
                 name: true,
                 shortName: true,
-                formName: true,
+                displayFormName: true,
             },
         },
         categoryOptionCombos: {
             id: true,
             name: true,
             shortName: true,
-            categoryOptions: true,
+            categoryOptions: {
+                id: true,
+                code: true,
+            },
         },
     },
 } as const;
@@ -83,7 +87,11 @@ function makeCocOrderArray(namesArray: string[][]): string[] {
 function getCocOrdered(categoryCombo: D2CategoryCombo, config: Dhis2DataStoreDataForm) {
     const allCategoryOptions = categoryCombo.categories
         .map(c => {
-            return c.categoryOptions.flatMap(co => ({ name: co.name, shortName: co.shortName, formName: co.formName }));
+            return c.categoryOptions.flatMap(co => ({
+                name: co.name,
+                shortName: co.shortName,
+                formName: co.displayFormName,
+            }));
         })
         .flatMap(categoriesOptions => {
             return categoriesOptions.map(co => co);
--- a/src/data/common/Dhis2DataElement.ts
+++ b/src/data/common/Dhis2DataElement.ts
@@ -1,7 +1,7 @@
 import _ from "lodash";
 import { Id } from "../../domain/common/entities/Base";
 import { DataElement } from "../../domain/common/entities/DataElement";
-import { D2Api, D2CategoryCombo, MetadataPick } from "../../types/d2-api";
+import { D2Api, MetadataPick } from "../../types/d2-api";
 import { promiseMap } from "../../utils/promises";
 import { Dhis2DataStoreDataForm } from "./Dhis2DataStoreDataForm";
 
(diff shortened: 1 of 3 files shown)


src/data/common/Dhis2DataElement.ts
```diff
@@ -48,16 +48,20 @@ const dataElementFields = {
             name: true,
             categoryOptions: {
                 id: true,
+                code: true,
                 name: true,
                 shortName: true,
-                formName: true,
+                displayFormName: true,
```
tokland, line 54: I tried to remove some fields here and still type-checked but there was a runtime error. That should not happened, so I searched for `as` castings. We have this `categoryOptionCombos: getCocOrdered(dataElement.categoryCombo as D2CategoryCombo, config)`. We should never do this kind of casting. It must be removed and the function defined as `function getCocOrdered(categoryCombo: D2DataElement["categoryCombo"]`. Now we can safely remove field; if unused, no TS error should appear.

src/data/common/Dhis2DataElement.ts
```diff
@@ -48,16 +48,20 @@ const dataElementFields = {
...
         },
         categoryOptionCombos: {
             id: true,
             name: true,
             shortName: true,
-            categoryOptions: true,
+            categoryOptions: {
+                id: true,
+                code: true,
+            },
```
tokland, line 64: Do we need this? after the previous fix in the type, it shows no error, so I guess it's not used. (the form loads normally)

src/data/common/Dhis2DataElement.ts
```diff
@@ -83,7 +87,11 @@ function makeCocOrderArray(namesArray: string[][]): string[] {
 function getCocOrdered(categoryCombo: D2CategoryCombo, config: Dhis2DataStoreDataForm) {
     const allCategoryOptions = categoryCombo.categories
         .map(c => {
-            return c.categoryOptions.flatMap(co => ({ name: co.name, shortName: co.shortName, formName: co.formName }));
+            return c.categoryOptions.flatMap(co => ({
+                name: co.name,
+                shortName: co.shortName,
+                formName: co.displayFormName,
```
tokland, line 93: should we use also the translated fields for name and shortName?

src/domain/common/entities/DataElement.ts
```diff
@@ -65,7 +65,12 @@ type CategoryCombos = {
     categoryOptionCombos: {
         id: Id;
         name: string;
+        formName: string | undefined;
         shortName: string | undefined;
+        categoryOptions?: {
+            id: Id;
+            code: string;
+        }[];
```
tokland, line 73: Not sure why we need these `categoryOptions?`. It's used in some view model, but where is being filled? If I remove the property and the usage, TS does not complain, so this should mean nobody is filling it.

tokland (commented):
Some comments not specifically about Eduardo's new code, but things I've detected while checking the feature. It may be tackled in another PR, if this one must go through.

src/domain/common/entities/DataElement.ts
tokland: Yes, but where is the property being set?

deeonwuli: it's filled in `Dhis2DataElement.ts` line 134

tokland: Can you paste the link? at https://github.com/EyeSeeTea/d2-autogen-forms/blob/feat/change-grid-with-cat-option-combos-colTitle-86944kn4z/src/data/common/Dhis2DataElement.ts#L134 I don't see the set.

src/data/common/Dhis2DataElement.ts
deeonwuli: we already have customizations in the autogen-config to display either formName or shortName (for the grid-with-combos) viewtype

src/domain/common/entities/DataElement.ts
tokland: > dumebi.  okay so it’s set on line 136: const categoryOptionCombos = dataElement.categoryCombo.categoryOptionCombos; the value of categoryOptions is gotten directly from the d2-response`

That's the point, thanks.

The general rule here is not to use optional properties unless strictly necessary.

MiquelAdell (approved):
I'm approving the PR as I need to generate a release.

Please @deeonwuli if there are still things pending raise a new PR for those. If not ask @tokland to validate the changes

PR #43, opened 2024-05-09
fix apply to all logic for source type column

eperedo:
### :pushpin: References

-   **Issue:** Closes #?

### :memo: Implementation

Right now the name is being used to validate if the column is "Source Type". This PR change that logic, now it'll will get that information from the dataStore: `dataElements->widget`

### :art: Screenshots

image

Build custom form

```bash
yarn build-autogenerated-forms
```


Ramon-Jimenez (approved):
Thanks @eperedo  , working great

PR #47, opened 2024-05-21
[Feature] Add basic entities tests 

delemaf:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/8694jf6w2
- 8694jf6w2

### :memo: Implementation

Added some basic tests for entities :
- `Config`
- `DataElement`
- `DataForm`
- `DataValue`
- `Indicator`
- `OrgUnit`
- `ToggleMultiple`
- `ValidationRule`

Run `yarn test`

### DIFF
--- a/src/domain/common/entities/DataValue.ts
+++ b/src/domain/common/entities/DataValue.ts
@@ -11,7 +11,7 @@ import {
     DataElementText,
 } from "./DataElement";
 
-interface DataValueBase {
+export interface DataValueBase {
     orgUnitId: Id;
     period: Period;
     categoryOptionComboId: Id;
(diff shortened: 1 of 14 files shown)


delemaf:
When test `DataValue.ts` I've notice that `DataValueStore.get` and `DataValueStore.getOrEmpty` have the same behaviour.
https://github.com/EyeSeeTea/d2-autogen-forms/blob/489a56bac74bd3632612f92cf9332a11c98198a9/src/domain/common/entities/DataValue.ts#L130-L143
Shouldn't `DataValueStore.get` return `undefined` instead of `getEmpty(dataElement, base)`? 
(In this case we should check if `get` and `getOrEmpty` are well used in code)

PR #49, opened 2024-05-22
chore: add github workflow

delemaf:
### :pushpin: References

- 8694jf6w2

### :memo: Implementation
GitHub workflow added:
- Setup node [16.14.0]
- Install yarn
- Install dependencies
- Run jest tests
- Run typescript tests


eperedo (changes requested):
Thanks @delemaf, code looks good, but not sure why are we removing the pull request template?

delemaf:
> Thanks @delemaf, code looks good, but not sure why are we removing the pull request template?

I think it was just a unused template created by default, should I re-add it?

PR #50, opened 2024-05-23
chore: test SaveDataFormValue use case

delemaf:
### :pushpin: References

- 8694jf6w2

### :memo: Implementation
Test SaveDataFormValue use cases :
- returns the store if the existing data value is equal to the new data value and type is not FILE
- updates the store with the new data value and saves it
- updates the store with the new FILE data value and saves it

Run `yarn test SaveDataFormValue.spec.ts` (to test only this file)

### DIFF
--- a/package.json
+++ b/package.json
@@ -130,6 +130,7 @@
         "prettier": "2.4.1",
         "react-test-renderer": "17.0.2",
         "ts-jest": "27.0.7",
+        "ts-mockito": "^2.6.1",
         "ts-node": "10.4.0",
         "typescript": "4.5.2",
         "wait-on": "6.0.0"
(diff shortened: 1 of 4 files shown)


delemaf:
I've mock DataValueStore in the first case but I don't know if it useful to check the other cases, let me know if I should add some verify

MiquelAdell:
> I've mock DataValueStore in the first case but I don't know if it useful to check the other cases, let me know if I should add some verify

I think this is a question that @eperedo will need to answer

Thanks!

eperedo (approved):
> I've mock DataValueStore in the first case but I don't know if it useful to check the other cases, let me know if I should add some verify

I think the test is ok since we don't really do anything with the store for that case. 

Thanks!

PR #52, opened 2024-05-24
chore: test GetOrgUnitsUseCase

delemaf:
### :pushpin: References

- 8694jf6w2

### :memo: Implementation
Test GetOrgUnitsUseCase use cases :
- calls the repository with correct parameters and returns the org units
- handles errors thrown by the repository

Run `yarn test GetOrgUnitsUseCase.spec.ts` (to test only this file)

### DIFF
--- a/package.json
+++ b/package.json
@@ -130,6 +130,7 @@
         "prettier": "2.4.1",
         "react-test-renderer": "17.0.2",
         "ts-jest": "27.0.7",
+        "ts-mockito": "^2.6.1",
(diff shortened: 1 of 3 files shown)


delemaf:
I don't think it's the most relevant test @eperedo , if it's not that useful we can pass on this use case

eperedo (approved):
Thanks for the code @delemaf. Yeah I agree we can skip this test since functionality is not complex.

PR #59, opened 2025-02-20
[fix]: Fix tab section sorting

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/86980quf9

### :memo: Implementation
- [x] Update `section.tabs.order` type to accept string or numeric values 
- [x] Fix sort order logic
- [x] Update type in autogen forms configurator

### :art: Screenshots

https://github.com/user-attachments/assets/9d0f50ca-9c08-468d-a468-1b406b5d076f

Using `tabs.order` with a number type automatically converts values such as `2.10` to `2.1`. By adding string type support, we can correctly get the primary tab sort index and the secondary section sort index and then sort accordingly.  

#86980quf9


MiquelAdell (approved):
tested and working, thanks!

src/webapp/reports/autogenerated-forms/SectionsTabs.tsx
```diff
@@ -94,21 +94,20 @@ function TypeSwitch(props: TypeSwitchProps) {
...
-function isTabHeader(order: number | undefined) {
+function isTabHeader(order: string | undefined) {
     if (order === undefined) return false;
-    if (order % 1 === 0) return true;
 
-    if (order === Math.floor(order) + 0.1) {
-        return true;
-    }
+    const [_primaryTabIndex, secondaryTabIndex] = order.split(".");
+    if (secondaryTabIndex === undefined || secondaryTabIndex === "0" || secondaryTabIndex === "1") return true;
```
eperedo: I think we can avoid the if condition here:

```ts
return secondaryTabIndex === undefined || secondaryTabIndex === "0" || secondaryTabIndex === "1"
```

src/data/common/Dhis2DataFormRepository.ts
```diff
@@ -84,84 +84,97 @@ export class Dhis2DataFormRepository implements DataFormRepository {
...
-                    return { viewType: config.viewType, ...base2 };
-            }
-        });
+                    default:
+                        return { viewType: config.viewType, ...base2 };
+                }
+            })
+            .sortBy([
+                section => (section.tabs.order ? parseInt(section.tabs.order?.split(".")[0] || "0", 10) : 0),
+                section => (section.tabs.order ? parseInt(section.tabs.order?.split(".")[1] || "0", 10) : 0),
```
eperedo: The second parameter in the parseInt method is already 10 for decimal numbers so for this case I think it's safe to remove it.

src/webapp/reports/autogenerated-forms/SectionsTabs.tsx
```diff
@@ -94,21 +94,20 @@ function TypeSwitch(props: TypeSwitchProps) {
...
 }
 
 const AutoFormComponent = React.memo(TypeSwitch);
 
 const TabPanel: React.FC<TabProps> = React.memo(props => {
     const { section, dataFormInfo, value } = props;
     const { viewType, tabs } = section;
-    const index = tabs.order !== undefined ? Math.floor(tabs.order) : -1;
+    const primaryTabIndex = tabs.order?.split(".")[0];
+    const index = tabs.order !== undefined && primaryTabIndex ? parseInt(primaryTabIndex) : -1;
```
eperedo: If `tabs.order` is undefined `primaryTabIndex` will be undefined as well, right? That means we don't need to check again if `tabs.order !== undefined` in the next line.

I think we can simplify to:

`const index = primaryTabIndex ? parseInt(primaryTabIndex) : -1;`

eperedo (changes requested):
Thanks @deeonwuli, let me know if you have some questions.

eperedo (approved):

PR #60, opened 2025-02-25
[feature]: Improve tab navigation

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/8697zg1x1

### :memo: Implementation
- [x] Added an overlay at the ends of the tab header
- [x] Increase the tab navigation arrow button sizes
- [x] Always show tab scroll buttons

### :art: Screenshots

https://github.com/user-attachments/assets/1d14107d-dbe7-4cf7-9446-c2ceaaf42daa

#8697zg1x1

### DIFF
--- a/src/webapp/reports/autogenerated-forms/SectionsTabs.tsx
+++ b/src/webapp/reports/autogenerated-forms/SectionsTabs.tsx
@@ -1,5 +1,5 @@
-import React from "react";
-import { Tabs, Tab, Box } from "@material-ui/core";
+import React, { useEffect, useRef, useState } from "react";
+import { Tabs, Tab, Box, makeStyles } from "@material-ui/core";
 import {
     Section,
     SectionWithPeriods,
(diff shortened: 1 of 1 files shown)


MiquelAdell:
Hi @deeonwuli the fading and the arrow size is great but we are missing one requested change. The arrows should not disapear but instead get opacity 0.5 or the equivalent grey color.

MiquelAdell (approved):
great, thanks!

src/webapp/reports/autogenerated-forms/SectionsTabs.tsx
```diff
@@ -117,23 +128,104 @@ const TabPanel: React.FC<TabProps> = React.memo(props => {
...
+                    handleScroll={handleScroll}
+                />
+                <ScrollButton
+                    direction="right"
+                    showLeftFade={showLeftFade}
+                    showRightFade={showRightFade}
+                    handleScroll={handleScroll}
+                />
+                <StyledTabs
+                    ref={tabsRef}
```
eperedo: Not sure if I'm missing something but looks like we're not doing anything with tabsRef, right?

src/webapp/reports/autogenerated-forms/SectionsTabs.tsx
```diff
@@ -117,23 +128,104 @@ const TabPanel: React.FC<TabProps> = React.memo(props => {
...
         <Box sx={{ width: "100%" }}>
             <StyledAppBar position="sticky" color="default">
-                <Tabs
+                <ScrollButton
+                    direction="left"
+                    showLeftFade={showLeftFade}
+                    showRightFade={showRightFade}
+                    handleScroll={handleScroll}
+                />
+                <ScrollButton
```
eperedo, line 213: The tabs component already include a prop to create custom buttons for the arrows: `ScrollButtonComponent`

https://v4.mui.com/api/tabs/#props
https://v4.mui.com/api/tab-scroll-button/#props
image
 
Are we creating the buttons from scratch because the prop was not flexible enough? If yes that's ok your implementation is correct.

eperedo (changes requested):
Thanks @deeonwuli! Don't hesitate to contact me if you have any questions about my comments.

src/webapp/reports/autogenerated-forms/SectionsTabs.tsx
deeonwuli: You're correct. the tabsRef is unnecessary

src/webapp/reports/autogenerated-forms/SectionsTabs.tsx
deeonwuli: Yes, the prop is not flexible enough. The left scroll button will always disappear when the user is scrolled to the left and one of the requirements of this task is to always show both scroll buttons.

eperedo (approved):

PR #61, opened 2025-02-28
enable input period validation if expiry days is 0

eperedo:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/86984thmk

### :memo: Implementation

- Right now all the inputs are enabled if expiryDays is equal to ZERO. I'm deleting that condition and always evaluate input periods.

### :art: Screenshots

Expiry days set to ZERO

image

2023 is disabled now:

[autogenform_expiry_days.webm](https://github.com/user-attachments/assets/bd9e4a50-3ba2-4daf-875b-c53777dd833a)

#86984thmk

### DIFF
--- a/src/webapp/reports/autogenerated-forms/DataEntryItem.tsx
+++ b/src/webapp/reports/autogenerated-forms/DataEntryItem.tsx
@@ -38,8 +38,6 @@ function isInputExpired(
     dataInputPeriods: dataInputPeriodsType,
     expiryDays: number
 ) {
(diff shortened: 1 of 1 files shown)


MiquelAdell (approved):

MiquelAdell:
I have tested it with a data set with no data input periods and expiryDays == 0 and it does not greyout fields. **Which is correct**. This means that this change does not mean an error for this kind of data sets.

PR #62, opened 2025-03-06
[feat]: Apply rules based on calculations

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/86980qxqu

### :memo: Implementation
- [x] Extend rules to support text values i.e. `Not sure`
- [x] Apply rules based on calculations to data elements and totals 
    Supported operators: 
      - Less than: <
      - Greater than: >
      - Less than or equal to: <=
      - Greater than or equal to: >=
      - Not equal: !=, !==
      - Equal: ==, === 
- [x] Toggle section visibility based on total calculated values

An example of rules configuration using the new calculations feature for data elements:
 ```
"DATA_ELEMENT": {
  "rules": {
    "visible": {
        "condition": "< 80",
       "type": "dataElements",
        "dataElements": [
            "DE_TO_HIDE"
        ]
    }
  }
}
```
This means that `DE_TO_HIDE` will only be visible if the value of `DATA_ELEMENT` is less than 80. We can apply this in a similar manner to the totals config:
```
"totals": {
  "dataElementCodes": ["DATA_ELEMENT_1", "DATA_ELEMENT_2"],
  "formulas": {
    "COLUMN_NAME": {
      "formula": "<%= Math.round((DATA_ELEMENT_1/DATA_ELEMENT_2)*100) %>",
      "type": "dataElements",
      "rules": {
        "visible": {
          "condition": "< 80",
          "dataElements": ["DE_TO_HIDE"]
        }
      }
    },
  }
}
```
where DE_TO_HIDE is only visible when the result of the totals formula is < 80. 

And to toggle section visibility, we have a `sections` type as well:
```
"totals": {
  "dataElementCodes": ["DATA_ELEMENT_1", "DATA_ELEMENT_2"],
  "formulas": {
    "COLUMN_NAME": {
      "formula": "<%= Math.round((DATA_ELEMENT_1/DATA_ELEMENT_2)*100) %>",
      "type": "sections",
      "rules": {
          "condition": "< 80",
          "sectionCodes": ["SECTION_TO_HIDE"]
      }
    },
  }
}
```
where SECTION_TO_HIDE is only visible when the result of the totals formula is < 80. 

### :art: Screenshots
https://github.com/user-attachments/assets/8fa204d5-0402-4e23-9e3f-9b7f20494445
https://github.com/user-attachments/assets/23a105bb-67f0-45fc-8525-244a8dbda43f

https://github.com/user-attachments/assets/6c05396e-4498-4997-b003-20b3cdbea488

#86980qxqu

### CODE REVIEW

File: src/data/common/Dhis2DataFormRepository.ts [hunk 1]
```diff
@@ -32,6 +32,7 @@ export class Dhis2DataFormRepository implements DataFormRepository {
         const dataElements = _.flatMap(sections, section => section.dataElements);
         const dataElementsOptions = this.getDataElementsOptions(dataElements, config);
         const dataSetConfig = config.getDataSetConfig(dataSet, options.period);
+        const totalRules: TotalRule[] = this.getTotalRules(sections, dataElements);
```
eperedo: Not a big deal, but since `getTotalRules()` already specifies the return type, explicitly setting the type in the variable declaration isn’t necessary.

```ts
const totalRules = this.getTotalRules(sections, dataElements);
```

src/data/common/Dhis2DataFormRepository.ts
```diff
@@ -164,6 +166,78 @@ export class Dhis2DataFormRepository implements DataFormRepository {
...
+            formula: formula,
+        };
+    }
+
+    private getRelatedDataElements(section: Section, dataElements: DataElement[], formula: string): DataElement[] {
+        const dataElementCodesInFormula = formula.match(/\b[A-Za-z_][A-Za-z0-9_]*\b/g) || [];
+
+        const relatedDataElements = _(dataElementCodesInFormula)
+            .map(dataElementCode => {
+                const dataElement = dataElements.find(de => de.code === dataElementCode);
```
eperedo: We can potentially have lots of dataElements in a dataSet. Let's create a `dataElementsByCode` to improve performance and avoid a nested loop: map -> find.

```ts
// Record<Code, DataElement> instead DataElements[]
const dataElementsByCode = _(dataElements).keyBy(de => de.code).value();

// inside the map

.map(dataElementCode => {  
  // we don't need the `find` loop anymore
  const dataElement = dataElementsByCode[dataElementCode];
  const isDataElementCodeIncluded = section.totals?.dataElementsCodes.includes(dataElementCode) ?? false;

  if (!dataElement || !isDataElementCodeIncluded) return undefined;

  return dataElements.find(de => de.id === dataElement.id);
})


```

src/data/common/Dhis2DataFormRepository.ts
```diff
@@ -164,6 +166,78 @@ export class Dhis2DataFormRepository implements DataFormRepository {
...
+                const isDataElementCodeIncluded = section.totals?.dataElementsCodes.includes(dataElementCode) ?? false;
+
+                if (!dataElement || !isDataElementCodeIncluded) return undefined;
+
+                return dataElements.find(de => de.id === dataElement.id);
```
eperedo: do we need this find again? I think we can return the `dataElement` directly, right?

src/data/common/Dhis2DataStoreDataForm.ts
```diff
@@ -39,7 +39,7 @@ interface BaseSectionConfig {
     disableComments: boolean;
     totals?: {
         dataElementsCodes: string[];
-        formulas: Record<string, { formula: string }> | undefined;
+        formulas: Record<string, { formula: string; rules: DataElementRuleOptions | undefined }> | undefined;
```
eperedo: Nothing wrong here, but since we already have the `Maybe` type let's use it here.

src/webapp/reports/autogenerated-forms/DataEntryItem.tsx
```diff
@@ -92,34 +95,123 @@ function checkVisibleRelatedDataElement(
...
-    const { dataElement, dataFormInfo, period } = options;
+export function checkVisibleRule(options: UseApplyRulesProps): boolean {
+    const { dataElement, dataFormInfo, period, totalRules } = options;
     const visibleRule = dataElement.rules.find(rule => rule.type === "visible");
-    const visibleValue = visibleRule ? getValueAndVerifyCondition(visibleRule, dataFormInfo, period) : true;
-    return visibleValue;
+    const visibleDataElementTotalRule = totalRules?.find(rule => rule.type === "visible");
+
+    return visibleRule || visibleDataElementTotalRule
+        ? getValueAndVerifyCondition(visibleRule, dataFormInfo, period, visibleDataElementTotalRule) ?? true
```
eperedo: We can remove the `?? true` since we already are returning that as a default value inside the method

eperedo (changes requested):
[code-only review] Thanks @deeonwuli, check my comments and let me know if you have any questions.

It's not related to the code you're working on, but could you please update the actions/cache version from 2 to 4 to avoid getting the error in the github action process. (main.yml file)

from `uses: actions/cache@v2` to `uses: actions/cache@v4`

src/webapp/reports/autogenerated-forms/DataEntryItem.tsx
deeonwuli: We're not returning the default value inside the method, but I have simplified the logic still.

src/data/common/Dhis2DataFormRepository.ts
```diff
@@ -164,6 +166,80 @@ export class Dhis2DataFormRepository implements DataFormRepository {
...
+            ),
+            relatedDataElements: relatedDataElements,
+            condition: rule.condition,
+            formula: formula,
+        };
+    }
+
+    private getRelatedDataElements(section: Section, dataElements: DataElement[], formula: string): DataElement[] {
+        const dataElementCodesInFormula = formula.match(/\b[A-Za-z_][A-Za-z0-9_]*\b/g) || [];
+        const dataElementsByCode = _(dataElements)
```
eperedo: Sorry I wasn't clear enough on this. This is correct, but can we do it outside the loop to avoid re-generating the dataElementsByCode for every section?

```ts
private getTotalRules(sections: Section[], dataElements: DataElement[]): TotalRule[] {
    const dataElementsByCode = _(dataElements)
          .keyBy(de => de.code)
          .value();
          
          // then pass Record<Code, DataElement> instead DataElement[] into the methods
}

```

eperedo (changes requested):

src/data/common/Dhis2DataFormRepository.ts
deeonwuli: oh i understand now. thanks Eduardo! :)

eperedo (approved):
Thanks @deeonwuli

MiquelAdell (approved):

PR #65, opened 2025-03-13
[feature]: Add support for multiple calculation rows

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/86988rwkx

### :memo: Implementation
- [x] Extend `totals` config to show multiple total/summary rows in `grid`, `grid-with-cat-option-combo` and `grid-with-periods` view types.

An example of the `totals` config:
```
{
  "totals": {
    "TOTALS_TEXT_1": {
      "dataElementsCodes": ["DATA_ELEMENT_1", "DATA_ELEMENT_2"],
      "formula": "<%= DATA_ELEMENT_1+DATA_ELEMENT_2 %>",
      "texts": {
        "name": "TOTAL_TEXTS_1"
      }
    },
    "TOTALS_TEXT_2": {
      "dataElementsCodes": [
        "DATA_ELEMENT_1",
        "DATA_ELEMENT_2",
        "DATA_ELEMENT_3"
      ],
      "formula": "<%= DATA_ELEMENT_1+DATA_ELEMENT_2-DATA_ELEMENT_3 %>",
      "texts": {
        "code": "TOTAL_TEXTS_2_CONSTANT"
      }
    }
  }
}
``` 
As seen above, the texts config for `totals` is moved into the totals object where both ordinary texts and constants are supported. 

### :art: Screenshots

https://github.com/user-attachments/assets/a5a0835a-72ce-4667-a6a9-892e39655f56

#86988rwkx


MiquelAdell (approved):
Tested on dev.est and dev.widp with "Total laboratory positive corrected"
 in section MAL_WMR_LAB and working correctly

src/data/common/Dhis2DataStoreDataForm.ts
```diff
@@ -585,16 +594,55 @@ export class Dhis2DataStoreDataForm {
...
+                    ...sectionTotals,
+                    texts: {
+                        name:
+                            this.getTextFromConstants({ code: sectionTotals.texts?.name || "" }, constantsByCode) || "",
+                    },
+                })),
+                (_, key) => constantsByCode[key]?.displayDescription || key
+            );
+        } else {
+            const sectionTotals = totals as SectionTotals;
```
eperedo: We can avoid the explicit "as" cast by checking if dataElementsCodes is in the object:

```ts
// at this point this check is enough because the `DataStoreConfigCodec.decode` already validated the schema
private isSectionTotals(config: TotalsConfig): config is SectionTotals {
        return "dataElementsCodes" in config;
}
```

then we can check:

```ts
if (this.isSectionTotals(totals)) {
  // here totals type is SectionTotals
} else {
 // here type is Record<string, SectionTotals>
}
```

and we don't need the "as" cast anymore, what do you think?

src/data/common/Dhis2DataStoreDataForm.ts
```diff
@@ -302,6 +306,16 @@ type CategoryCombinationConfig = {
     viewType: "name" | "shortName" | "formName" | undefined;
 };
 
+type TotalsText =
```
eperedo: Nothing wrong with the current implementation, but the `totals->text` object for every section was used because we only had one total, now that we support multiple totals I think this could not be the right place for this configuration, but the new object `Record<string, SectionTotals>`

Before
```ts
"[SEC_CODE]": {
  "texts": {
    "totals": { "code": "constant_code" }
  },
  "totals": {
    // texts.totals was used here because sections only can have one total
    "formula": ""
    // other properties
  }
}
```

Now
```ts
"[SEC_CODE]": {
  // we don't need texts.totals anymore because we can set a constant code for each total config.
  "totals": {
     "TOTAL_ID_1": {
             // we can have
             "texts": { name: "CONSTANT_CODE OR PLAIN TEXT" },
              "formula": ""
             // other properties
     },
     "TOTAL_ID_2": {
             "texts": { name: "CONSTANT_CODE OR PLAIN TEXT" },
              "formula": ""
             // other properties
     },     
  }
}
```

Again nothing wrong with the current implementation but it's a little bit weird that we have different schemas in the same section:

```json
{
"texts": {
  "header": {
    "code": "MAL-WMR-PRESUMED-SEC-SUBTITLE"
  },
  "rowTotals": {
    "code": "MAL_WMR_LABORATORY_TOTAL_COLUMN"
  },
    "totals": []
    // can be object { code: string }, but also array ["CODE_1", "CODE_2"]
}
}
```

src/data/common/Dhis2DataStoreDataForm.ts
```diff
@@ -585,16 +594,55 @@ export class Dhis2DataStoreDataForm {
...
+        constantsByCode: Record<string, Constant>
+    ): Record<string, SectionTotals> | undefined {
+        const { totals, texts } = sectionConfig;
+        if (!totals || !texts?.totals) return undefined;
+
+        if (_.isPlainObject(totals) && Object.values(totals).every(value => _.isPlainObject(value))) {
+            const sectionTotalsMap = totals as Record<string, SectionTotals>;
+
+            return _.mapKeys(
+                _.mapValues(sectionTotalsMap, sectionTotals => ({
```
eperedo: This is subjective, but I think if more easier to understand if you use map + fromPairs.

```ts
_(sectionTotalsMap) // map in lodash can iterate objects
.map((section, key) => {
   // first parameter is the value, second the key
    const constantValue = this.getTextFromConstants({ code: key }, constantsByCode);
    // Since you're using the content of the constant as key throw if value is empty
    if (!constantValue) throw new Error("Cannot find value for constant");
    return [
        constantValue,
        {
            ...section,
            texts: {
                name:
                    this.getTextFromConstants({ code: section.texts?.name || "" }, constantsByCode) ||
                    "",
            },
        },
    ] as [string, SectionTotals];
})
.fromPairs()
.value();
```

eperedo (changes requested):
Thanks @deeonwuli, let me know if you have any questions.

src/data/common/Dhis2DataStoreDataForm.ts
deeonwuli: I think this is a great suggestion. Thanks Eduardo! 😃

src/data/common/Dhis2DataStoreDataForm.ts
```diff
@@ -24,6 +24,12 @@ export type SectionConfig =
     | GridWithTotalsSectionConfig
     | GridWithSubnationalSectionConfig;
 
+type SectionTotals = Totals & {
+    texts?: { name: string | { code: string } };
```
eperedo: can we use a similar format like we have in `texts.totals` by just adding a new code property instead having it inside the name property?

```json
{
 "texts": { "name": "not translated text", "code":  "constant_code" }
 // we first check for the name, if nothing here we use code
}
```

```ts
{ texts?: { name?: string, code?: string }

// for the codec could be:

texts: optional(Codec.interface({ name: optional(string), code: optional(string) })),
```

Then we look for constants codes directly in `totals.code`

```ts
const totalsCodes = _(storeConfig.dataSets)
    .values()
    .flatMap(dataSet => _.values(dataSet.sections))
    .flatMap(section => {
        if (!section.totals) return undefined;
        return this.isSectionTotals(section.totals)
            ? [section.totals.texts]
            : _(section.totals)
                    .map(total => total.texts)
                    .value();
    })
    .map(totals => totals?.code)
    .compact()
    .value();
```

then logic for new syntax should consider first `texts.name` and fallback to `texts.code`

```ts
return _(totals)
    .map((sectionTotals, key) => {
        const constantCodeOrValue = sectionTotals.texts?.name || { code: sectionTotals.texts?.code ?? "" };
        const constantValue = this.getTextFromConstants(constantCodeOrValue, constantsByCode) ?? key;

        return [
            constantValue,
            {
                ...sectionTotals,
                texts: {
                    name: this.getTextFromConstants(constantCodeOrValue, constantsByCode) || "",
                },
            },
        ] as [string, SectionTotals];
    })
    .fromPairs()
    .value();
```

and now we can avoid the syntax `{ name: { code: "value"} }`

eperedo (changes requested):
Thanks @deeonwuli looks good, we're very close just a minor observation.

eperedo (approved):
Looks good to me, thanks @deeonwuli !

PR #69, opened 2025-03-28
[fix]: Hide calculation rows when there are no total items

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/8698gfqde

### :memo: Implementation
- [x] Hide totals rows when 
  - the category option combos are different
  - when there are no total items

### :art: Screenshots
- Diagnosis by RDT/Microscopy by health sector `MAL_WMR_LAB` has no total rows as a fall back due to the inconsistencies in the category option combos of the data elements in that section
- Double counted figures `MAL_WMR_LAB_DBL_COUNT_GRID` has total rows according to the config (just for testing)

https://github.com/user-attachments/assets/b36b3ce4-ef49-4fbd-a892-e4220acb36a1

#8698gfqde


src/webapp/reports/autogenerated-forms/GridFormViewModel.ts
```diff
@@ -203,6 +193,31 @@ export class GridViewModel {
             }),
         };
     }
+
+    private static getColumnWithDataElements(selectedDataElements: DataElement[], column: Column): TotalItem[] {
```
eperedo: Nice abstraction.

src/webapp/reports/autogenerated-forms/GridWithPeriodsViewModel.ts
```diff
@@ -213,6 +204,31 @@ export class GridWithPeriodsViewModel {
         };
     }
 
+    private static getColumnWithDataElements(selectedDataElements: DataElement[], column: string): TotalItem[] {
```
eperedo: Is this almost the same method that we have in `GridFormViewModel.ts`, right? We're only using the column param for the error message, so I think we can reuse this method in both `GridFormViewModel.ts` and `GridWithPeriodsViewModel`

```ts
private static getColumnWithDataElements(selectedDataElements: DataElement[], columnName: string): TotalItem[] {
   // rest of the code
   
   if (!categoryOptionCombo)
       throw new Error(
         `Cannot found categoryOptionCombo in column ${columnName} for dataElement ${dataElement.code}`
       );
}
```

And we pass the name property instead the whole object

```ts
const columnWithDataElements = GridViewModel.getColumnWithDataElements(
  selectedDataElements,
  column.name
);
```

Does this make sense?

eperedo (changes requested):

eperedo (approved):
Looks good thanks!

PR #71, opened 2025-04-21
[feature] Add deploy script

MatiasArriola:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/8698p9d1w #8698p9d1w

### :memo: Implementation

- Add a `yarn deploy` script to publish the forms generated by `yarn build-autogenerated-forms`
- Publish to a single dataSet by id
  - Errors if the dataSet has no form, or if the form is not a d2-autogen-form.
  - Use `--force` to deploy anyway.
- Publish to multiple dataSets that are currently using d2-autogen-forms if no ID is provided
  - Lists all dataSets using d2-autogen-forms and asks for confirmation to deploy the form to all of them
  - Use `-y` / `--yes` to automatically answer the prompt
- Allows customizing d2-autogen-form detection by using `-q` /  `--query`


MiquelAdell (approved):

src/scripts/deploy.ts
```diff
@@ -0,0 +1,249 @@
...
+    }
+    console.debug("Publishing to all dataSets that use d2-autogen-forms...");
+    for (const dataSet of dataSets) {
+        await updateDataEntryForm(api, dataSet, contentToUpload);
+    }
+    console.debug(`✓ Published d2-autogen-forms to ${dataSets.length} datasets`);
+}
+
+function readFormToUpload(): Promise<string> {
+    const FILE_TO_PUBLISH = "dist/custom-data-form.html";
```
eperedo: We have this path in `src/scripts/build-autogenerated-forms.ts#41`. Could we centralize this value into a shared variable and reuse it across both files? This would prevent inconsistencies if we ever need to modify it later..

eperedo (changes requested):
Just a minor observation. Thanks @MatiasArriola

MatiasArriola:
Thanks @eperedo! Extracted the hardcoded path to `scripts/common.ts`

eperedo (approved):
Looks good to me, thanks @MatiasArriola!

PR #75, opened 2025-07-25
[feature]: Rules with multiple conditions based on options

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/8699rgz5y

### :memo: Implementation
- [x] Add support for rules with conditions based on data element options:
```
"rules": {
  "disabled": {
    "conditions": [
      {
        "condition": "OPTION_CODE_1",
        "dataElements": ["DATA_ELEMENT_1", "DATA_ELEMENT_2"]
      },
      {
        "condition": "OPTION_CODE_2",
        "dataElements": ["DATA_ELEMENT_1"]
      }
    ],
    "type": "option"
}
```
where `OPTION_CODE_1` and `OPTION_CODE_2` are the values of the option set option codes. 

### :art: Screenshots

https://github.com/user-attachments/assets/73cedafa-d8a4-4a99-8572-9ecc690ee314

#8699rgz5y

### DIFF
--- a/src/data/common/Dhis2DataFormRepository.ts
+++ b/src/data/common/Dhis2DataFormRepository.ts
@@ -299,11 +299,23 @@ export class Dhis2DataFormRepository implements DataFormRepository {
         };
     }
 
-    private getTotalDataElements(rule: ConditionRule, dataElements: Record<string, DataElement>) {
+    private getTotalDataElements(rule: ConditionRule, dataElements: Record<string, DataElement>): Id[] {
         const extractedDataElements =
             rule.type === "option" ? rule.conditions.flatMap(condition => condition.dataElements) : rule.dataElements;
 
-        return extractedDataElements.map(dataElementCode => dataElements[dataElementCode]?.id ?? "");
+        return _(extractedDataElements)
+            .map(dataElementCode => {
+                const dataElement = dataElements[dataElementCode];
+                if (!dataElement) {
+                    console.warn(`Data element not found for code: ${dataElementCode}`);
+                    return;
+                }
+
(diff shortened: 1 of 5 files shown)


src/data/common/Dhis2DataFormRepository.ts
```diff
@@ -292,13 +291,21 @@ export class Dhis2DataFormRepository implements DataFormRepository {
...
+                rule.type === "option" ? rule.conditions.map(condition => condition.condition) : [rule.condition],
             formula: formula,
         };
     }
 
+    private getTotalDataElements(rule: ConditionRule, dataElements: Record<string, DataElement>) {
+        const extractedDataElements =
+            rule.type === "option" ? rule.conditions.flatMap(condition => condition.dataElements) : rule.dataElements;
+
+        return extractedDataElements.map(dataElementCode => dataElements[dataElementCode]?.id ?? "");
```
eperedo: Since we're getting the id's here, can we rename the name to `getTotalDataElementIds`? And including the type should be useful here: 

```private getTotalDataElementIds(rule: ConditionRule, dataElements: Record<string, DataElement>): Id[]```

Is safe for this method to return an empty string?  I think we can either throw an error or at least log a warning to console and ignore empty strings.

eperedo (changes requested):

eperedo (changes requested):
I think I already did a review for this code here: https://github.com/EyeSeeTea/d2-autogen-forms/pull/76#pullrequestreview-3149578404 @deeonwuli

eperedo (approved):

PR #77, opened 2025-08-12
[feature]: Config to hide category options

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/8699y2dhn 

### :memo: Implementation
Datastore config:
```
"categoryOptions": {
    "CATEGORY_OPTION_CODE_1": {"visible": false},
    "CATEGORY_OPTION_CODE_2": {"visible": false}
  },
```

#8699y2dhn

### DIFF
--- a/src/data/common/Dhis2DataElement.ts
+++ b/src/data/common/Dhis2DataElement.ts
@@ -93,7 +93,8 @@ function getCocOrdered(categoryCombo: D2DataElement["categoryCombo"], config: Dh
     const allCategoryOptions = categoryCombo.categories.flatMap(c =>
         _(c.categoryOptions)
             .flatMap(co => {
-                if (config.categoryOptionsConfig[co.code]?.visible === false) return undefined;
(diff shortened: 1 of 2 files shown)


src/data/common/Dhis2DataElement.ts
```diff
@@ -90,21 +90,30 @@ function makeCocOrderArray(namesArray: string[][]): string[] {
...
+            })
+            .compact()
+            .value()
+    );
 
     const categoryOptionsNamesArray = categoryCombo.categories.map(c => {
-        return c.categoryOptions.flatMap(co => co.name);
+        return _(c.categoryOptions)
+            .flatMap(co => {
+                if (config.categoryOptionsConfig[co.code]?.visible === false) return undefined;
```
eperedo: We have the same logic twice in different parts, I know is a very simple condition, but let's extract into a method please. Just to avoid repeating the same logic in other parts of the repo.

eperedo (changes requested):

eperedo (approved):

PR #81, opened 2025-08-28
[feature]: Disable section using toggle config

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869a9kr3f

### :memo: Implementation
- [x] Optionally disable a section using toggle data element (existing behaviour only hides a section using toggle).
The config for this feature is as follows:
```
"SECTION_CODE": {
  "toggle": {
    "code": "DE_CODE",
    "type": "dataElement",
    "disabled": "true"
  },
},
```
where the `disabled` property is optional.

### :art: Screenshots

https://github.com/user-attachments/assets/9d45c91c-150f-413e-8ce6-b6e1175fc613

#869a9kr3f

### DIFF
--- a/src/data/common/Dhis2DataStoreDataForm.ts
+++ b/src/data/common/Dhis2DataStoreDataForm.ts
@@ -41,12 +41,24 @@ type SectionTotals = Totals & {
 
 type TotalsConfig = SectionTotals | Record<string, SectionTotals>;
 
+type DataElementToggle = {
+    type: "dataElement";
+    code: Code;
+    disabled: boolean;
+};
+
+type DataElementExternalToggle = {
+    type: "dataElementExternal";
+    code: Code;
+    condition: string | undefined;
+    disabled: boolean;
+};
+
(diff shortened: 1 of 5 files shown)


src/data/common/Dhis2DataStoreDataForm.ts
```diff
@@ -222,6 +234,7 @@ const DataStoreConfigCodec = Codec.interface({
                         type: oneOf([exactly("dataElement"), exactly("dataElementExternal")]),
                         code: string,
                         condition: optional(string),
+                        disabled: optional(string),
```
eperedo: is there any reason we're using a string instead boolean here?

eperedo (changes requested):

src/data/common/Dhis2DataStoreDataForm.ts
deeonwuli: No reason in particular. I tested using a boolean type and it doesn't change the functionality. 👌

eperedo (approved):
Looks good, thanks @deeonwuli

PR #83, opened 2025-09-02
fix indicator calculation

eperedo:
### :pushpin: References

-   **Issue:** Closes #?

### :memo: Implementation

An indicator formula can include the dataElement Id and the combination Id:

```shell
#{DE_ID.COC_ID} + #{DE_2_ID.COC_ID}
``` 

For this case the current solution works, but you can omit the combination id in the formula:

```shell
#{DE_ID} + #{DE_2_ID}
``` 

This case was not covered in the current solution. Now the flow is:

- Extract dataElement Id from formula 
- Extract cocId and fallback to current cocId [assigned to the dataElemen](https://github.com/EyeSeeTea/d2-autogen-forms/blob/master/src/domain/common/entities/DataElement.ts#L25)t if not present in formula. 
- Get dataValue associated to dataElement+cocId
- Execute the calculation


ifoche (approved):
thanks @eperedo

PR #85, opened 2025-09-10
Fix: ensure comments/audit use correct period across DHIS2 handlers

ifoche:
Summary

  - Fixes comment and audit history actions sometimes using the wrong period in multi-period tables.
  - Handles both code paths: when DHIS2 reads via dhis2.de.getSelectedPeriod() and when it reads directly from globals/DOM.

  Context

  - Some DHIS2 handlers fetch the selected period via dhis2.de.getSelectedPeriod().
  - Others read it directly from dhis2.de fields or the hidden input used by legacy code.
  - Result: the comment/audit action could be executed with the wrong period for the clicked cell.
  
  Changes

  - src/webapp/reports/autogenerated-forms/CommentIcon.tsx
  - src/webapp/reports/autogenerated-forms/DataElementItem.tsx
  - On click/double-click:
      - Temporarily override dhis2.de.getSelectedPeriod() to return ({ iso: period }).
      - Also set dhis2.de.currentPeriod, currentPeriodId, currentPeriodIso.
      - If available, call dhis2.de.setSelectedPeriod({ iso }).
      - Update hidden input #selectedPeriodId if present.
      - Restore originals immediately after the handler runs.

  Why this approach

  - Covers both “API” and “direct JS” period resolution paths mentioned in debugging.
  - Limits scope to the specific user interaction and restores state immediately to avoid side effects.
  - No behavior change outside Data Entry context (window.dhis2?.de guard).

  Verification

  - In Data Entry, open a form with multi-period columns.
  - Comments:
      - Click the comment icon for several period columns.
      - Verify the dialog shows and saves comments for the correct period.
  - Audit history:
      - Double-click the cell to open history (or the relevant action).
      - Verify the history corresponds to the clicked period.
  - Sanity check single-period forms to ensure no regressions.
  
  Risk/Impact

  - Low. Only runs when a period prop is provided and window.dhis2?.de exists.
  - State changes are temporary and restored after the event loop tick.

  Notes

  - If the target DHIS2 build uses a different DOM id or additional globals for period, we can extend the shim easily (e.g., other hidden inputs/selectors).

### DIFF
--- a/src/webapp/reports/autogenerated-forms/CommentIcon.tsx
+++ b/src/webapp/reports/autogenerated-forms/CommentIcon.tsx
@@ -6,13 +6,65 @@ import { useAppContext } from "../../contexts/app-context";
 export interface CommentIconProps {
     dataElementId: Id;
     categoryOptionComboId: Id;
+    period?: string;
 }
 
 export const CommentIcon: React.FC<CommentIconProps> = React.memo(props => {
-    const { dataElementId, categoryOptionComboId } = props;
+    const { dataElementId, categoryOptionComboId, period } = props;
     const { api } = useAppContext();
     const tagId = `${dataElementId}-${categoryOptionComboId}-comment`;
     const title = i18n.t("View comment and audit history");
+    
+    const onClickCapture = React.useCallback(() => {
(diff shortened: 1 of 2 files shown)


ifoche:
closing as it has been already solved in a different PR by @eperedo

PR #88, opened 2025-09-22
[feature]: Define periods for cat-option-combo sections

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869ahn9en

### :memo: Implementation
- [x] Add support for `periods` config in "grid-with-cat-option-combos" view type
```
"viewType": "grid-with-cat-option-combos",
"periods": {
  "endOffset": -1,
  "startOffset": -1,
  "type": "relative-interval"
}
```
Single and multiple periods are supported. 

### :art: Screenshots

https://github.com/user-attachments/assets/1a1333e3-3d1c-4dfb-97d0-547191aa1791

#869ahn9en


src/webapp/reports/autogenerated-forms/DataEntryItem.tsx
```diff
@@ -31,14 +31,17 @@ export interface DataEntryItemProps {
     columnDataElements?: DataElement[];
     cocId: string;
     rows?: Row[];
+    lockException?: boolean; // If true, the input will not be disabled if expired
```
eperedo: Not sure if I'm missing something but if the period is expired and we force an input to be enabled user can't enter data due to an exception from the API, right? 

Not wrong, just to be sure I understand how this works.

src/domain/common/entities/DataForm.ts
```diff
@@ -94,11 +94,11 @@ export interface SectionBase {
...
 
 export interface SectionSimple extends SectionBase {
-    viewType: "grid-with-combos" | "grid-with-cat-option-combos" | "matrix-grid" | "grid-disaggregated-cocs";
+    viewType: "grid-with-combos" | "matrix-grid" | "grid-disaggregated-cocs";
 }
 
 export interface SectionWithPeriods extends SectionBase {
-    viewType: "grid-with-periods";
+    viewType: "grid-with-periods" | "grid-with-cat-option-combos";
     periods: string[];
```
eperedo, line 102: We have a `getReferencedPeriods` method in this class:

```ts
static getReferencedPeriods(dataForm: DataForm, basePeriod: Period): Period[] {
        return _(dataForm.sections)
            .flatMap(section => {
                switch (section.viewType) {
                    case "grid-with-periods":
                        return section.periods;
                    case "grid-indicators-calculated":
                        return section.periods;
                    default:
                        return [];
                }
            })
            .uniq()
            .concat([basePeriod])
            .sortBy()
            .value();
    }
```

This is being used to fetch dataValues including all the periods, can we check if we need to include the `"grid-with-cat-option-combos"`, here?

eperedo (changes requested):
Looks good @deeonwuli, just minor things about fetching dataValues for all the periods and the new `lockException` prop for inputs

Also can you share an example to test this on dev, please? I could not found any config. in the dataStore for this dataSet. Thanks!

MiquelAdell (approved):

src/domain/common/entities/DataForm.ts
deeonwuli: yes, we would need to add `"grid-with-cat-option-combos"` here 👍

src/webapp/reports/autogenerated-forms/DataEntryItem.tsx
deeonwuli: yes that is the usual behaviour. I think it would be better I rewrite the comment to explain the behaviour of `lockException` here. In the TUB dataset, the previous periods are not locked by the API so this will work.

I think a better comment would be:
`// If true, the input will be not disabled when the period is not the same as the data form period`

how about that @eperedo?

eperedo (approved):

PR #91, opened 2025-10-09
[feature]: Add summary row to table form

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/8699mq839

### :memo: Implementation
- [x] Refactor function to get summary according to view type in grid form view model
- [x] Add summary rows to table form

### :art: Screenshots
Screenshot 2025-10-09 at 12 44 11

#8699mq839


src/webapp/reports/autogenerated-forms/GridFormViewModel.ts
```diff
@@ -52,10 +55,42 @@ interface Row {
 const separator = " - ";
 
 export class GridViewModel {
-    static get(section: SectionGrid, dataFormInfo: DataFormInfo): Grid {
-        const dataElementsConfig = dataFormInfo.metadata.dataForm.options.dataElements;
+    static get(section: SectionGrid, dataFormInfo: DataFormInfo, viewType?: SectionGrid["viewType"]): Grid {
```
eperedo: As this method is only used in two places, we can remove the optional parameter to make it clear that the viewType affects the logic.

src/webapp/reports/autogenerated-forms/GridFormViewModel.ts
```diff
@@ -132,66 +188,75 @@ export class GridViewModel {
...
+                                {
+                                    columnName: key,
+                                    formula: sectionTotal.formula || "",
+                                    items: this.getColumnWithDataElements(selectedDataElements, key),
+                                },
+                            ],
+                        };
+                    })
+                    .value();
+            default:
```
eperedo: Same here — let’s add a case for `grid`, and for the default, return an empty array and a `console.warn(\Unsupported viewType ${viewType}`) or something similar. I know it’s very unlikely we’ll have another implementation with this viewModel, but just to be extra safe.

eperedo (changes requested):

eperedo (approved):

MiquelAdell (approved):

PR #92, opened 2025-10-09
[feature]: Change indicator before/after position in grid-with-cat-option-combo

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/8699mq839

### :memo: Implementation
- [x] Display indicator name as fallback when indicator description is empty
- [x] Change indicator display in GridWithCatOptionCombos view model

### :art: Screenshots

https://github.com/user-attachments/assets/ec599806-e8a8-4eef-9d68-cc777a83d1f3

- All the indicators used in the config must have a `code`

#8699mq839


src/webapp/reports/autogenerated-forms/GridWithCatOptionCombos.tsx
```diff
@@ -145,6 +169,20 @@ const GridWithCatOptionCombos: React.FC<GridWithCatOptionCombosProps> = props =>
                                         ></CustomDataTableCell>
                                     );
                                 })}
+
+                                {section.indicators.map(
+                                    indicator =>
+                                        indicator.dataElement?.code === row.dataElement.code &&
+                                        checkIndicatorDirection(indicator, "after") && (
```
eperedo: Not wrong, but maybe could be easier to read if we move the logic to a variable and render the result?

```ts
const filteredIndicators = section.indicators.filter(indicator =>
    indicator.dataElement?.code === row.dataElement.code &&
    checkIndicatorDirection(indicator, "after")
);

// then we render filteredIndicators
{filteredIndicators.map(indicator => (
    <IndicatorItem
        key={`${indicator.id}-${row.dataElement.code}`}
        indicator={indicator}
        dataFormInfo={dataFormInfo}
        periods={[dataFormInfo.period]}
    />
))}
```

What do you think?

src/webapp/reports/autogenerated-forms/GridWithCatOptionCombos.tsx
```diff
@@ -184,18 +223,6 @@ const GridWithCatOptionCombos: React.FC<GridWithCatOptionCombosProps> = props =>
                             )}
                         </DataTableRow>
                     ))}
-
-                    {section.indicators.map(indicator => {
```
eperedo, line 206: is this going to generate a breaking change? Because I remember we were using this viewType in the WMR Form dataSet and not sure if with these changes we're changing some existing functionality.

eperedo (changes requested):

src/webapp/reports/autogenerated-forms/GridWithCatOptionCombos.tsx
deeonwuli: This isn't exactly the same logic. On rows where there is no indicator matching the row's data element, the ideal behaviour is to have no indicator shown. But it makes a lot of sense to extract the logic to make it more readable

src/webapp/reports/autogenerated-forms/GridWithCatOptionCombos.tsx
deeonwuli: I am cross checking jst to make sure. I know we are using indicators in the WMR form but I do not see any indicator metadata in the dataset.

deeonwuli: I will add the indicators back. But only those without specified direction. This will ensure backward compatibility

eperedo (approved):

PR #93, opened 2025-10-14
[feature]: Config to display indexes before data elements

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869arbukw

### :memo: Implementation
- [x] Add config `showIndex` to dataset and section config:
```
{
  "DATASET_CODE": {
    "SECTION_CODE": {
      ...
      "showIndex": false,
      ...
    },
    "showIndex": true,
  }
}
```
`showIndex` is effective in the following view types:
- grid with disaggregated cocs
- grid 
- grid with cat option combos
- grid with periods

### :art: Screenshots

https://github.com/user-attachments/assets/ad400128-d1f1-41f5-8c8a-a3c70f47bf41

#869arbukw


src/webapp/reports/autogenerated-forms/DataTableSection.tsx
```diff
@@ -133,6 +140,18 @@ const useStyles = makeStyles({
     },
 });
 
+export function getIndexedLabel(
+    section: { tabs?: { active: boolean; order?: string }; showIndex?: boolean },
+    name: string,
+    dataElementIndex?: number
```
eperedo: Since we only have one place where we don't need this value, let's remove the optional parameter and use Maybe<T> so is more explicit to understand what's the logic behind it.

eperedo (changes requested):
[code review] Thanks @deeonwuli. Just a minor comment about the optional parameter used in the `getIndexedLabel` function.

MiquelAdell:
I think by default we are showing indexes? Can we make sure that the default is false?

deeonwuli:
> I think by default we are showing indexes? Can we make sure that the default is false?

hi @MiquelAdell ! by default, the indexes are not shown

eperedo (approved):

PR #95, opened 2025-11-05
[feature]: Add orgUnit toggle functionality to config

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869b0r4mx

### :memo: Implementation
- [x] Introduced an org unit toggle type with the conditions:
   - `hide`: hide the section/given data elements for the given ous only
   - `show`: hide the section/given data elements for all ous except the selected ous
```
{
  "SECTION_CODE": {
    "toggle": {
      "type": "orgUnit", 
      "orgUnits": ["TOGGLE_OU_CODE_1", "TOGGLE_OU_CODE_2"],
      "dataElements": ["DE_CODE_1", "DE_CODE_2"], // or undefined
      "condition": "show", // or "hide",
      "disabled": true, // or false or undefined
    }
  }
}
```
In this config, 
- if the `dataElements` property is defined, only those data elements in the section will be hidden (or shown) for the toggle org units 
- if the `dataElements` property is not defined, the entire section will be hidden (or shown)

#869b0r4mx


src/data/common/Dhis2DataStoreDataForm.ts
```diff
@@ -837,15 +855,40 @@ export class Dhis2DataStoreDataForm {
...
-            disabled?: boolean;
-        }>;
+        toggle: Maybe<
+            | {
+                  type: "dataElement" | "dataElementExternal";
+                  code: string;
+                  condition: Maybe<string>;
+                  disabled?: boolean;
+              }
+            | {
```
eperedo: We can use the GetType from codec to create the type from the definition:

```ts
type OrgUnitToggle = GetType<typeof orgUnitToggleCodec>;
```

src/domain/common/usecases/GetDataFormUseCase.ts
```diff
@@ -1,15 +1,23 @@
...
-        return this.dataFormRepository.get({
+    async execute(options: { dataSetId: Id; period: Period; orgUnitId: Id }): Promise<DataFormWithOU> {
+        const orgUnit = await this.orgUnitsRepository.getById(options.orgUnitId);
+        const dataForm = await this.dataFormRepository.get({
             id: options.dataSetId,
             period: options.period,
             orgUnitId: options.orgUnitId,
         });
+
+        return {
```
eperedo, line 24: Maybe I'm missing something, but since we will hide/show the dataElements/sections based on the current org unit  (the one selected in the data entry) we should apply the filter here instead the UI layer, right?

src/webapp/reports/autogenerated-forms/hooks/useSectionVisibility.ts
```diff
@@ -0,0 +1,121 @@
...
+import { DataTableSectionObj } from "../DataTableSection";
+
+export type SectionVisibilityState = {
+    isDisabled: boolean;
+    isOpen: boolean;
+    isVisible: boolean;
+};
+
+export function useSectionVisibility(section: DataTableSectionObj, dataFormInfo: DataFormInfo): SectionVisibilityState {
+    const { toggle, toggleMultiple } = section;
```
eperedo, line 16: Check my comment in `GetDataFormUseCase.ts` just to be sure if we're filtering sections in the right place.

eperedo (changes requested):

src/domain/common/usecases/GetDataFormUseCase.ts
deeonwuli: You're right. The filtering should be in the use case 👍

eperedo (approved):
Thanks @deeonwuli

PR #96, opened 2025-11-17
[feature]: Format section periods by period type

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869b2dgzm

### :memo: Implementation
- [x] Get section periods based on dataset periodtype

### :art: Screenshots

https://github.com/user-attachments/assets/3a437a63-bd0d-4970-af6d-9ab2ddad03b4

#869b2dgzm


src/domain/common/entities/Period.ts
```diff
@@ -0,0 +1,25 @@
...
+export function validatePeriodType(periodType: string): PeriodType {
+    switch (periodType) {
+        case PeriodType.DAILY:
+        case PeriodType.WEEKLY:
+        case PeriodType.MONTHLY:
+        case PeriodType.QUARTERLY:
+        case PeriodType.YEARLY:
+            return periodType;
+        default:
+            throw new Error(`Invalid period type: ${periodType}`);
```
eperedo: I think this could be considered a breaking change, because existing dataSets with other periodTypes (like FinancialApril) work for `viewTypes` where we don't generate multiple periods (like table or grid) and with this change we're throwing this error. Can we add an "unknown" `periodType` instead of throwing the error? 

I don't remember other clients using a different period type than "YEARLY", but just to be extra safe.

Does this make sense?

src/data/common/Dhis2DataStoreDataForm.ts
```diff
@@ -732,10 +834,11 @@ export class Dhis2DataStoreDataForm {
         return sectionValue ?? dataSetValue ?? false;
     }
 
-    getDataSetConfig(dataSet: DataSet, period: Period): DataSetConfig {
+    getDataSetConfig(dataSet: DataSet, period: Id): DataSetConfig {
         const dataSetConfig = this.config.custom.dataSets?.[dataSet.code];
         const dataSetDefaultViewType = dataSetConfig?.viewType || defaultViewType;
         const constantsByCode = _.keyBy(this.config.constants, getCode);
+        const periodType = validatePeriodType(dataSet.periodType);
```
eperedo, line 866: check my comment in the `Period.ts` entity

eperedo (changes requested):

eperedo (approved):

MiquelAdell (approved):

PR #98, opened 2025-11-25
[feature]: Period offset for sections

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869b7xknm

### :memo: Implementation
In `table` and `grid-with-cat-option-combos` view types, we can now add a period column to the section. For this, we introduced a new `section-offset` period type:
```
{
  periods: {
    type: "section-offset",
    offset: -1 // number to offset by
  }
}
```

### :art: Screenshots

https://github.com/user-attachments/assets/be385ca0-6ed0-4e23-b3f5-eb7addb7e8f1

#869b7xknm


MiquelAdell (approved):

src/webapp/reports/autogenerated-forms/TableForm.tsx
```diff
@@ -41,11 +42,26 @@ const TableForm: React.FC<TableFormProps> = React.memo(props => {
                         {!_.isEmpty(section.columns) && (
                             <CustomDataTableColumnHeader
                                 backgroundColor={props.section.styles.columns.backgroundColor}
-                                colSpan="2"
+                                colSpan={section.dataEntryPeriod ? 1 : 2}
```
eperedo: Very minor, but we're not getting the error from `dhis2/ui` here, but this parameter is actually supposed to receive an string and not number. Can we update it please so we avoid the warning in the console.

image

src/webapp/reports/autogenerated-forms/TableForm.tsx
```diff
@@ -41,11 +42,26 @@ const TableForm: React.FC<TableFormProps> = React.memo(props => {
...
+                            <>
+                                <CustomDataTableColumnHeader
+                                    backgroundColor={props.section.styles.columns.backgroundColor}
+                                    width="50px"
+                                >
+                                    {i18n.t("Period")}
+                                </CustomDataTableColumnHeader>
+                                <CustomDataTableColumnHeader
+                                    backgroundColor={props.section.styles.columns.backgroundColor}
+                                    colSpan={1}
```
eperedo: Same here, string instead of number

eperedo (changes requested):
It looks like right now this is the default behaviour and I think this could be a breaking change. I've created a dataSet with a section with this config:

```json
{
  "dataSets": {
    "0boolean_test": {
      "sections": {
        "bol_test": {
          "viewType": "table"
        }
      }
    }
  }
}
```

 and I still get the Period column (even though I omitted the period configuration)

image

eperedo (approved):

PR #100, opened 2025-12-01
[fix]: Issues with disaggreated cocs view type

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869bak8au

### :memo: Implementation
- [x] Reorder columns in view type
- [x] Fix so that category option is correctly matched to the cell

### :art: Screenshots

https://github.com/user-attachments/assets/4be3e512-0d1d-417d-9e4d-16e8151fa90a

#869bak8au


src/webapp/reports/autogenerated-forms/DisaggregatedCOCsGridViewModel.ts
```diff
@@ -185,16 +185,34 @@ type Summary = {
...
-    return _(items)
-        .sortBy(item => {
-            const match = item.name.match(/^(\d+)(?:\D|$)/);
-            const numericValue = match ? match[1] : undefined;
-            const parsedNumericValue = numericValue ? parseInt(numericValue, 10) : undefined;
-
-            return [parsedNumericValue === undefined ? 1 : 0, parsedNumericValue ?? item.name.toLowerCase()];
-        })
-        .value();
+const sortItems = <T extends { name: string }>(items: T[], unknownPatterns: string[] = ["unknown", "other"]): T[] => {
```
eperedo, line 237: I think the `unknownPatterns` value should be part of the configuration in the dataStore. For this client this logic could be valid, but thinking in a more generic implementation the rules might be different for other dataSets. 

at the section level maybe something like:

```json
{
  "sortConfig": {
     "patterns": ["unknown", "other"]
  }
}
```

If we don't have the time to implement it can you please add some comment explaining the reason behind these values? Something like: "Other/unknown will be always at the end. This logic is specific for the dataSet ABC" (just an example feel free to add all the details you consider necessary)

eperedo (changes requested):
[code-review] Looks good, just a minor observation:

src/webapp/reports/autogenerated-forms/DisaggregatedCOCsGridViewModel.ts
deeonwuli: Confirmed with @MiquelAdell and I'm adding a comment just for now

MiquelAdell (approved):

eperedo (approved):

PR #103, opened 2025-12-18
[feature]: Add email support in TextWidget and DataElement types

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869bhdf6g

### :memo: Implementation
- [x] Add support for email data element types 

### :art: Screenshots

https://github.com/user-attachments/assets/0bf3b065-e317-4fc2-82e6-b300ce52fdb4

#869bhdf6g

### DIFF
--- a/src/webapp/reports/autogenerated-forms/widgets/TextWidget.tsx
+++ b/src/webapp/reports/autogenerated-forms/widgets/TextWidget.tsx
@@ -5,6 +5,7 @@ import { WidgetFeedback } from "../WidgetFeedback";
 import { DataValueTextSingle } from "../../../../domain/common/entities/DataValue";
 import { WidgetProps } from "./WidgetBase";
 import i18n from "../../../../locales";
+import { Maybe } from "../../../../utils/ts-utils";
 
 export interface TextWidgetProps extends WidgetProps {
     dataValue: DataValueTextSingle;
@@ -14,7 +15,7 @@ const TextWidget: React.FC<TextWidgetProps> = props => {
     const { onValueChange, dataValue, disabled } = props;
 
     const [stateValue, setStateValue] = React.useState(dataValue.value);
-    const [emailError, setEmailError] = React.useState<string | undefined>();
+    const [emailError, setEmailError] = React.useState<Maybe<string>>();
     React.useEffect(() => setStateValue(dataValue.value), [dataValue.value]);
 
     const updateState = React.useCallback(
@@ -64,8 +65,11 @@ const TextWidget: React.FC<TextWidgetProps> = props => {
 
 export default React.memo(TextWidget);
 
+// Note: This regex validation is not the same as DHIS2's backend validation.
+// For example, john@doe.coma is valid for this regex but invalid for DHIS2.
+// This is acceptable since we still get the error from DHIS2 backend.
 const validateEmail = (email: string) => {
     if (!email) return true;
     const emailRegex = /^[^\s@]+@[^\s@]+\.[^\s@]+$/;
     return emailRegex.test(email);
-}
+};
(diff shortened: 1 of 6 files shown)


src/webapp/reports/autogenerated-forms/widgets/TextWidget.tsx
```diff
@@ -13,27 +13,55 @@ const TextWidget: React.FC<TextWidgetProps> = props => {
...
+        const emailRegex = /^[^\s@]+@[^\s@]+\.[^\s@]+$/;
+        return emailRegex.test(email);
     }, []);
 
+    const updateState = React.useCallback(
+        ({ value }: { value: string }) => {
+            setStateValue(value);
+            if (dataValue.dataElement.isEmail && value) {
+                if (!validateEmail(value)) {
+                    setEmailError("Please enter a valid email address");
```
eperedo: i18n here so we can have translations

src/webapp/reports/autogenerated-forms/widgets/TextWidget.tsx
```diff
@@ -13,27 +13,55 @@ const TextWidget: React.FC<TextWidgetProps> = props => {
     const { onValueChange, dataValue, disabled } = props;
 
     const [stateValue, setStateValue] = React.useState(dataValue.value);
+    const [emailError, setEmailError] = React.useState<string | undefined>();
     React.useEffect(() => setStateValue(dataValue.value), [dataValue.value]);
 
-    const updateState = React.useCallback(({ value }: { value: string }) => {
-        setStateValue(value);
+    const validateEmail = React.useCallback((email: string) => {
```
eperedo: Can we move this function outside the component? Right now there's no need to check for changes in every re-render.

src/webapp/reports/autogenerated-forms/widgets/TextWidget.tsx
```diff
@@ -13,27 +13,55 @@ const TextWidget: React.FC<TextWidgetProps> = props => {
...
+        if (!email) return true;
+        const emailRegex = /^[^\s@]+@[^\s@]+\.[^\s@]+$/;
```
eperedo: Nothing to change here just as a note: this regex is not the same validation dhis is using on the backend. 

Eg: john@doe.coma is valid for this regex, but invalid for dhis2

I think this is not a big deal since we still get the error from DHIS2:

image

Also replicating the same email validation could involved a lot more code.

eperedo (changes requested):
Looks good, just some minor details

src/webapp/reports/autogenerated-forms/widgets/TextWidget.tsx
deeonwuli: Noted @eperedo. Would we need to add a comment here?

eperedo: Yes, good idea. The main reason is that they're using an external library that involves a more complex validation and it could be more difficult to replicate in the client side like the TLD validation.

In case you want more details about the validation here's the validation in the core repo: 

https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-support/dhis-support-system/src/main/java/org/hisp/dhis/system/util/ValidationUtils.java#L199

And this is the library they're using:

https://commons.apache.org/proper/commons-validator/apidocs/src-html/org/apache/commons/validator/routines/EmailValidator.html#line-175

deeonwuli: Thank you!

eperedo (approved):

MiquelAdell (approved):

PR #113, opened 2026-01-14
[fix]: Render comment image during development

MatiasArriola:
### :pushpin: References

-   **Issue:** Closes N/A

### :memo: Implementation

Avoid broken comment image during development by proxying it to a local copy of the image


MiquelAdell (commented):
approving this directly as it is very simple

PR #117, opened 2026-01-16
[fix]: Update indices when showIndex is false

deeonwuli:
### :pushpin: References

-   **Issue:** Closes 

### :memo: Implementation
- [x] Update `getDataElementLabel` and `getIndexedLabel` to adjust tab indices when `showIndex` is alse

### :art: Screenshots
Screenshot 2026-01-16 at 21 53 51
where `TUB_TEST_HEADING` has this config:

```
"TUB_TEST_HEADING": {
  "showIndex": false,
  "tabs": {
    "active": true,
    "order": "18.9"
  }
}

```

PR #118, opened 2026-01-20
fix: add indexes also to data elements with a name replace

MiquelAdell:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869bv3dq9

### :art: Screenshots
Screenshot 2026-01-20 at 10 10 24

created with codex


MatiasArriola (approved):
Looks good to me. Code-wise change looks safe.

For the future there is room for refactoring to avoid code duplication.

This change replicates something similar to what was implemented in `GridFormViewModel.ts`.

MatiasArriola:
Take into consideration there is some overlap with the PR #117 

Maybe that PR also fixes the same issue

PR #123, opened 2026-01-22
[feature]: delete hidden data

gqcorneby:
### :pushpin: References

-   **Issue:** Closes #869bcdxd0 https://app.clickup.com/t/869bcdxd0

### :memo: Implementation
- add delete  rules in data and domain models + data repo implementation
    - `DataElement.deleteRules` is a separate prop from `rules` because evaluation is on the DE of interest instead of the target DE
- had to update the input of some widgets (use `value` instead of `defaultValue`) because they weren't reactive to store updates
- updated use cases that update the dataStore to return updated values instead. This way dataStore state is updated as needed instead of overwritten per prop. Chaining requests revealed a bug where stale values where used.
- updated schema for better autocomplete on rules
- add delete rule condition special value wildcard (`*`) do rule an be applied on change regardless of value 

### :art: Screenshots

https://github.com/user-attachments/assets/f361cdf1-cfc0-4c05-8395-1ba147066062

http://localhost:8081/?dataSet=WA6lLxXAACn&orgUnit=H8RixfF8ugH&period=2025


MiquelAdell (approved):

src/webapp/reports/autogenerated-forms/hooks/useDeleteRule.ts
```diff
@@ -0,0 +1,56 @@
...
+                                dataValue =>
+                                    dataValue.dataElement.code === deId &&
+                                    dataValue.orgUnitId === dataFormInfo.orgUnitId
+                            )
+                            .compact()
+                            .value()
+                    );
+                } else return [];
+            });
+            await runDelete(allDataValuesToDelete);
```
eperedo: Can we handle the exception here? Logging it to the console should be enough for debugging if this happens on data entry.

```ts
await runDelete(allDataValuesToDelete).catch((err) => {
    console.error("Error applying delete rules", err);
    setDeleteRuleInProgress(false);
});
```

src/data/common/Dhis2DataFormRepository.ts
```diff
@@ -540,6 +544,48 @@ export class Dhis2DataFormRepository implements DataFormRepository {
...
+    ): Record<string, DeleteRule[]> {
+        return _(dataElements)
+            .map(dataElement => {
+                const dataElementConfig = configDataForm.dataElementsConfig[dataElement.code];
+                if (!dataElementConfig) return undefined;
+
+                const deleteRulesConfig = dataElementConfig.rules?.delete;
+                if (!deleteRulesConfig) return undefined;
+
+                const rules: DeleteRule[] = this.buildDeleteRules(deleteRulesConfig);
```
eperedo: I think is safe to remove the type here since it could be infer from the method? 

```ts
const rules: DeleteRule[] = this.buildDeleteRules(deleteRulesConfig);
```

Also maybe we don't need the `as const` if we add the type to the map:

```ts
.map((dataElement): Maybe<[string, DeleteRule[]]> => {
    const dataElementConfig = configDataForm.dataElementsConfig[dataElement.code];
    if (!dataElementConfig) return undefined;

    const deleteRulesConfig = dataElementConfig.rules?.delete;
    if (!deleteRulesConfig) return undefined;

    const rules = this.buildDeleteRules(deleteRulesConfig);
    return [dataElement.code, rules];
})
```

src/data/common/Dhis2DataFormRepository.ts
```diff
@@ -540,6 +544,48 @@ export class Dhis2DataFormRepository implements DataFormRepository {
...
+            })
+            .compact()
+            .fromPairs()
+            .value();
+    }
+
+    private buildDeleteRules(deleteRulesConfig: ConditionRule): DeleteRule[] {
+        switch (deleteRulesConfig.type) {
+            case "option":
+                return deleteRulesConfig.conditions.map((rule: any) => ({
```
eperedo: `any` is not necessary here because `conditions` is already an array of `SingleConditionRule`, right?

src/data/common/Dhis2DataValueRepository.ts
```diff
@@ -367,6 +367,18 @@ export class Dhis2DataValueRepository implements DataValueRepository {
         }
     }
 
+    async delete(dataValues: DataValue[]) {
```
eperedo: Not a big deal, but adding the return type just to know what's the expected value without going to the repository.

src/webapp/reports/autogenerated-forms/hooks/useDeleteRule.ts
```diff
@@ -0,0 +1,56 @@
...
+    const applyDeleteRule = React.useCallback(
+        async (dataValueCb: DataValue) => {
+            if (dataElement.deleteRules.length === 0) return;
+            setDeleteRuleInProgress(true);
+
+            const runDelete = dataFormInfo.data.delete;
+
+            const allDataValuesToDelete = dataElement.deleteRules.flatMap(rule => {
+                const shouldDelete = verifyConditionByDataValueType(dataValueCb, rule);
+                if (shouldDelete) {
```
eperedo: Subjective but I think in this case an early return could be easier to read.

```ts
if (!shouldDelete) return [];

...
```

src/webapp/reports/autogenerated-forms/hooks/useInputState.ts
```diff
@@ -0,0 +1,13 @@
+import React from "react";
+
+export function useInputState<TInputValue>(srcValue: TInputValue) {
+    const [value, setValue] = React.useState(srcValue);
+
+    React.useEffect(() => setValue(srcValue), [srcValue]);
+    const onChange = React.useCallback(
+        (e: React.ChangeEvent<{ value: unknown }>) => setValue(e.target.value as TInputValue),
```
eperedo: Since we're using a standard input can we use `React.ChangeEvent<HTMLInputElement>` instead of `{value: unknown}`?

eperedo (changes requested):
Thanks @gqcorneby, I've left some comments, let me know if you have any questions.

src/webapp/reports/autogenerated-forms/hooks/useDeleteRule.ts
gqcorneby: used `finally` instead

gqcorneby:
Hi @eperedo! Aside from the changes to address the comments, a new "clear" rule is introduced so we can delete the data element's data based on it's disabled state. This is to simplify how multiple enabled rules can affect the same data element.

src/data/common/Dhis2DataFormRepository.ts
eperedo: can we add the type here `Maybe<[string, DeleteRule[]]>`? Just to be safe because `fromPairs` returns a `Dictionary<any>`

src/domain/common/entities/DataElementRule.ts
```diff
@@ -14,9 +14,20 @@ type MultipleConditionRule = {
     conditions: SingleConditionRule[];
 };
 
-export type ConditionRule = SingleConditionRule | MultipleConditionRule;
+export const STATE_CONDITIONS = {
+    disabled: "disabled",
+};
```
eperedo: This should be `as const`? To avoid TS accepting any string:

```ts
const a: StateConditionRule = {
    type: "state",
    // valid
    condition: "anything_here",
}
```

with `as const`

```ts
export const STATE_CONDITIONS = {
    disabled: "disabled",
} as const;

const a: StateConditionRule = {
    type: "state",
    // Type '"anything_here"' is not assignable to type '"disabled"'.
    condition: "anything_here",
}
```

eperedo (changes requested):

src/domain/common/entities/DataElementRule.ts
gqcorneby: Right, thanks!

eperedo (approved):

PR #126, opened 2026-01-22
[fix]: show tab index if any sibling section has showIndex

MatiasArriola:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869brgqw8 #869brgqw8

### :memo: Implementation

- Previously, the tab number would only show if the "tab section" had the configuration `showIndex: true`. With this change, it also shows the number if any section belonging to the same tab, has the `showIndex` property.

This is to fix the case where the first section of a tab is decorative and has showIndex false, but we still want to show the tab index.

Please review if this workaround is valid. 

Another alternative I thought about was to add an extra configuration option to `[section].tabs`, because here the same `[section].showIndex` property  is indicating both: "include the index in the tab name" and "include the index for section indexing (subtitles)".


MiquelAdell (approved):
approving this and merging it directly as the change is very simple

PR #129, opened 2026-01-26
Fix/tableform indicator period offset (copy from PR #128)

MiquelAdell:

### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869bwm272 #869bwm272

### :memo: Implementation

When an indicator was shown in a `TableForm` with a period offset, the input would show under period. This adds an extra `colSpan` in case there is a period offset, so that the input is aligned.

### :art: Screenshots

imagen


MatiasArriola:
Closing in favor of #133

PR #130, opened 2026-01-26
[feature]: dataset custom cssFeat/dataset custom css (copy from #127)

MiquelAdell:

### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869bvqf9b #869bvqf9b

### :memo: Implementation

- Add a `customCss` property in to the dataStore configuration, under the dataSet.
- Add `data-autogen-form-id={dataForm.id}` attribute to the autogenerated form container
- If `customCss` is available then insert a new stylesheet into the head, which is scoped to the autogenerated form using the `@scope` css rule. 
- A best-effort sanitization is made using `CSSStyleSheet`, to avoid including broken CSS or escaping the container. There are some alternatives like parsing the css, but we are trying to keep it simple

This feature poses a security risk and also incorrect styles might break the form, but we trust the admins in charge of setting the config.


src/webapp/reports/autogenerated-forms/hooks/useDataFormCustomCss.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    try {
+        // @ts-ignore `replace` is supported in modern browsers but not in current types
+        await sheet.replace(cssText);
+    } catch {
+        console.warn("Failed to parse custom CSS.", { cssText });
+        return null;
+    }
+
+    let sanitized = "";
```
gqcorneby, line 15: minor suggestion - could use `.map().join()` to follow functional programming

`const sanitized = Array.from(sheet.cssRules)
.map(rule => rule.cssText)
.join("\n")
.trim();`

gqcorneby (approved):
[code-only review]  Approving. One minor suggestion about the functional approach.

PR #125, opened 2026-01-22
[fix]: missing indicators

MatiasArriola:
copied here:
https://github.com/EyeSeeTea/d2-autogen-forms/pull/125

### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869bw4gej #869bw4gej

### :memo: Implementation

- revert changes made in #122 back to using `section` instead of `grid` for indicators, which was causing indicators to not be rendered

PR #132, opened 2026-01-26
[fix]: missing indicators (copy from #125)

MiquelAdell:

### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869bw4gej #869bw4gej

### :memo: Implementation

- revert changes made in #122 back to using `section` instead of `grid` for indicators, which was causing indicators to not be rendered


MatiasArriola:
Closing in favor of #134

PR #134, opened 2026-01-27
[fix]: indicators indexes and indicator visibility

MatiasArriola:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869bw4gej #869bw4gej

### :memo: Implementation

This PR completes the work and fixes some issues from #122

- Fix indicator indexes adding the proper offset (previously they would always start at 1, accounting only for indicators).
- Move indicators logic to the viewModel (`hasIndicatorsBefore`, `hasIndicatorsAfter`,`nonDirectionalIndicators`), indicators.before and after for each row. 
  - As a side effect fixes logic bug that caused "after" indicators to not show because of the usage of `section.indicators` vs `grid.indicators` (now we refer to the viewModel instead of the entity). Also minor perf improvement.
- Fix: do not prefix indicators if the section does not have the  `showIndex` flag enabled.
- offtopic: Add the viewType next to the section code in the debug info of each section

### :art: Screenshots

imagen


MiquelAdell (approved):

src/webapp/reports/autogenerated-forms/GridWithCatOptionCombosViewModel.ts
```diff
@@ -147,18 +165,21 @@ export class GridWithCatOptionCombosViewModel {
...
                         const deLabel = getDataElementLabel(de, section, dataFormInfo, de.name);
-                        const row = {
-                            dataElement: { ...de, htmlText: getDataElementHtmlText(de, section, dataFormInfo) },
+                        const dataElement = { ...de, htmlText: getDataElementHtmlText(de, section, dataFormInfo) };
+                        const row: SubRow = {
+                            dataElement: dataElement,
                             deName: deLabel,
                             name: de.name,
+                            indicators: {
+                                before: getFilteredIndicators(section.indicators, dataElement, "before"),
```
gqcorneby, line 187: [minor optimization] `getFilteredIndicators` is called twice per row, doing O(rows × indicators × 2) filtering total. We could build a lookup dictionary once:


[not tested]
```const indicatorsByKey: Record<string, Indicator[]> = _.groupBy(
section.indicators,
i => `${i.dataElement?.code}_${i.dataElement?.direction}`
);
```

Then replace `getFilteredIndicators` calls with:
```indicators: {
        before: indicatorsByKey[`${dataElement.code}_before`] ?? [],
        after: indicatorsByKey[`${dataElement.code}_after`] ?? [],
}
```


This reduces complexity from O(rows × indicators × 2) to O(indicators + rows).

gqcorneby (approved):
[code-only review]  Approving. The optimization suggestion is relatively minor and non-blocking. A similar approach would be beneficial in `GridViewModel` (especially with large datasets) but would be more complex and take a hit on readability. Noting it down just in case.

MiquelAdell:
keeping this open as per geonina comment and also some conflicts to be resolved against 1.4.0

src/webapp/reports/autogenerated-forms/GridWithCatOptionCombosViewModel.ts
MatiasArriola: Notice that this PR is not introducing these calls. It is moving them to inside the viewModel generation, which hopefully is memoized most of the renders. Previously they were inlined in the render method. we got a small gain there. 

I agree with your observation, we could better optimize the code in the viewmodels while making it more readable, but I think it would be nice to have some separate task for that

MatiasArriola:
Closing as these changes were already merged with #122

PR #135, opened 2026-01-28
[feature]: add new config to section - indicatorsPosition

MatiasArriola:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869brgqw8 #869brgqw8

### :memo: Implementation

The task required some indicators to be shown before dataElements in some particular sections. Indicators were being rendered at the end for all viewTypes.
To approach this, a new dataStore config property `indicatorsPosition` was added. Possible values are `start` or `end`, defaulting to `end` (this keeps backwards compat). 
Also, indexing logic for indicators and dataElements was modified to take into account this new property.

IMPORTANT: this PR is based on the #134 branch, so we would need to review that first.


MiquelAdell (approved):

src/data/common/Dhis2DataFormRepository.ts
```diff
@@ -172,6 +172,7 @@ export class Dhis2DataFormRepository implements DataFormRepository {
                     toggleMultiple: config?.toggleMultiple
                         ? buildToggleMultiple(config.toggleMultiple, section, dataElements)
                         : undefined,
+                    indicatorsPosition: config?.indicatorsPosition || "end",
```
gqcorneby: [nitpicking] The default `"end"` is defined in multiple places (`Dhis2DataFormRepository.ts`, `Dhis2DataStoreDataForm.ts`). Consider extracting to a shared constant like `defaultPeriodType = "end"` to ensure consistency if this ever needs to change.

src/webapp/reports/autogenerated-forms/GridFormViewModel.ts
```diff
@@ -289,16 +308,26 @@ export class GridViewModel {
             const groupMeta = lastPartSubSection && groupsByRow ? groupsByRow[lastPartSubSection] : undefined;
             const rowName = groupsByRow ? lastPartSubSection ?? "" : subsection.name;
 
+            const nonDirectionalIndicatorsCount =
```
gqcorneby: This code snippet appears multiple times, maybe we can extract it to a shared utility function

```function getNonDirectionalIndicatorsOffset(section: SectionBase): number {
    return section.indicatorsPosition === "start"
        ? section.indicators.filter(isNonDirectionalIndicator).length
        : 0;
}
```

gqcorneby (approved):
[code-only review] Approving. I just have some minor comments that are non-blocking.

MiquelAdell:
keeping this open as per geonina comment and also some conflicts to be resolved against 1.4.0

MatiasArriola:
Thanks @gqcorneby! Refactored to improve code reuse, good catch.

Also merged release into this branch to solve conflicts

PR #137, opened 2026-01-30
[feature]: add hide flag to rowsConfig

MatiasArriola:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869bh2gj5 #869bh2gj5

### :memo: Implementation

- This change is motivated by the need to hide an specific row for the `grid-disaggregated-cocs` viewType
- There was an existing `rowsConfig` dataStore config, so a new `hide` optional flag was added there.
- `rowsConfig` was not used in `grid-disaggregated-cocs`, so the mapping was added, and added support for `hide` and `rowNameConstant`.  
- Also added `hide` support for `grid-category-columns` (not tested, but change looks safe)


MiquelAdell (approved):

src/domain/common/entities/DataForm.ts
```diff
@@ -113,6 +113,7 @@ export interface SectionBase {
     fixedRowNames: boolean;
     hidden?: boolean;
     columnsConfig?: Record<string, { rules?: RulesFormula }>;
+    rowsConfig?: Maybe<RowConfig>;
```
gqcorneby: I don't have full context on this, but have you considered creating a separate 
type for `"grid-disaggregated-cocs"` and extracting it from `SectionSimple` 
instead of adding `rowsConfig` as an optional prop in `SectionBase`? 

This would make the type more explicit about which view types support rowsConfig. Something like
```
type SectionDisaggregatedCocs = SectionBase & {
  viewType: "grid-disaggregated-cocs";
  rowsConfig: Maybe<RowConfig>;
  // ... other specific props
}
```

gqcorneby (commented):
[code-only review] Hi @MatiasArriola! Most of it looks good, just some minor comment on the entity so we're more explicit. 

Additionally, just noting here but maybe we can do it in a different task, I noticed that the schema for the editor is way outdated. I tried updating it when I worked on the delete rule but I just realized now that it's quite outdated 😅

src/domain/common/entities/DataForm.ts
MatiasArriola: Very good point :ok_hand: 

Created the specific type for `SectionDisaggregatedCocs`

gqcorneby (approved):
Looks good!

PR #138, opened 2026-01-30
[fix]: use period offset in totals calculation for GridWithCatOptionCombos

MatiasArriola:
### :pushpin: References

-   **Issue:** Related https://app.clickup.com/t/869brj0y2 #869brj0y2

### :memo: Implementation

- I've added a totals configuration `grid-with-cat-option-combos` which contained a `periods` config with a -1 offset and realized it was not supported: calculations were based on current period and the columns would be misaligned because previous rows contained the "period" column.
- This change fixes this situation in particular for `GridWithCatOptionCombos` by passing around the period for the summary, when there is only one period.
- Also pass the row period to the totals column (not needed for this task but somewhat related)

IMPORTANT: 

I've noticed the periods config allow for a `startOffset` and `endOffset`. In case there is more than 1 period, the behavior will be the same as before (showing one summary row at the end, using the current period). 

We could implement calculating the summary for each period, and showing it right below each period change, but that would involve a larger change and functional definitions.


gqcorneby (approved):
Looks good! Thanks @MatiasArriola!

MiquelAdell (approved):

PR #139, opened 2026-02-02
[feature]: mirror data element

MatiasArriola:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869brj0y2 #869brj0y2

### :memo: Implementation

We needed to show the same dataElement in different sections. Since it is not allowed, we came with this approach. 
Create a second "stub" dataElement, and include in the dataStore a config like this: 

```json
"dataElements": {
  "DataElement_Code_readOnly": {
      "mirrorFrom": "DataElement_Code"
  }
}
```

The field for the dataElement that contains the mirrorFrom property will keep the original value updated, but will be disabled.

This is built from #138 branch, so please review that PR first (here we need to pass the Period around as well).

offtopic:
- move `indicatorIndexing.ts` to utils folder


MiquelAdell (approved):

src/webapp/reports/autogenerated-forms/utils/resolveDataElement.ts
```diff
@@ -0,0 +1,18 @@
+import { DataElement } from "../../../../domain/common/entities/DataElement";
+import { DataFormInfo } from "../AutogeneratedForm";
+
+/**
+ * If dataElement contains mirrorFrom, resolves and returns the referenced data element from dataFormInfo.
+ * Otherwise, returns the original data element.
+ */
+export function resolveDataElement(dataFormInfo: DataFormInfo, dataElement: DataElement): DataElement {
```
gqcorneby, line 8: [nitpicking] Might be subjective but it might also be useful to  consider renaming `resolveDataElement` to something like `resolveMirroredDataElement` or something to make it clearer that this function specifically handles the mirror reference resolution. The current name is a bit generic and doesn't hint at what's being 'resolved'

gqcorneby (approved):
Cool workaround, @MatiasArriola! 

I know I don't have all the context but one thought about the feature in general, just thinking out loud (non-blocking as usual):
The DE used to mirror will basically remain empty right? And acts more like a "placeholder"? If I understood correctly, if this kind of feature is used a lot, the dataSet might end up having a lot of clutter to mirror values. 

I vaguely remember a similar feature mentioned by Eduardo during a quick demo :thinking: He eventually pointed me to this DataSet `MNCAH_DS_CONSOLIDATED`.

Not 100% sure if it's a viable alternative and I'm sure you guys might have considered this already, but I'll note it here just in case we want to revisit it in the future. It looks like we have "virtual" rows and columns in `grid-indicators-calculated` that acts similarly but is only used for this specific viewType. We could potentially extend that pattern here to avoid the clutter.

src/webapp/reports/autogenerated-forms/utils/resolveDataElement.ts
MatiasArriola: Yes, good observation. Trying to find a name for this function, and in which file to place it, was what I took me most of the time :rofl: 
I finally went with this generic name because it won't always return the mirror, it could return itself, and included the comment instead.

MatiasArriola:
Thanks @gqcorneby, your analysis is on-point. The mirrored DE is only a placeholder, will remain empty and it is only a support for this workaround.
We evaluated using `grid-indicators-calculated` but it did not fit our use-case (here we have one grid with cat option combos with totals and period offset).

I considered and really liked the idea of having virtual dataElements, but I went this path only to deliver faster. I think it would be nice if we could get some time to revisit this, design and implement the "virtual" alternative, and evaluate to remove this "mirrorFrom" feature.

gqcorneby:
> Thanks @gqcorneby, your analysis is on-point. The mirrored DE is only a placeholder, will remain empty and it is only a support for this workaround. We evaluated using `grid-indicators-calculated` but it did not fit our use-case (here we have one grid with cat option combos with totals and period offset).
> 
> I considered and really liked the idea of having virtual dataElements, but I went this path only to deliver faster. I think it would be nice if we could get some time to revisit this, design and implement the "virtual" alternative, and evaluate to remove this "mirrorFrom" feature.

Yeah, I figured the workaround was faster and does seem relatively more straight forward to configure as well :) We can revisit it if the need for this feature comes up more than we'd like and the clutter becomes an issue. Thanks @MatiasArriola!

PR #142, opened 2026-02-03
[feature]: empty values in totals calculations, strict flag

MatiasArriola:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869brj0y2 #869brj0y2

### :memo: Implementation

- DEPENDS ON #139 (based on that branch since I required previous features)
- Add the flag `strict` to the totals dataStore config. 
  - when `strict` is `true`, then dataValues will not be automatically converted to 0 when empty.
  - this was to not break backwards compatibility in case there are preexisting formulas that rely on the automatic conversion from `''` to `0`.
  - We could analyze if this flag is required and remove it and change the default behavior if needed.
- In `getValueFromDataElement`, instead of always converting empty values to 0, return `undefined` if the value is empty and `strict` is true.
- In `DataTableCellSummary` handle the possibility of formula evaluations returning `NaN`. Show an empty value if all previous values are empty or NaN.
- The change to `getValueFromDataElement` will impact `getItemTotalDataValue` `DataTableCellFormula`. As formula values will be `undefined` instead of `0`, calculations might start returning `NaN`, but that is handled in `DataTableCellFormula` `CustomInput` value parameter.
- Also modified `DataTableCellRowTotal` to be consistent with this behavior. If the values to the left are not set, show empty instead of 0 (this is a change despite the `strict` flag setting)

I'm was not totally convinced about this PR, but looks like an acceptable workaround without too many changes. 
The flag makes it more verbose but was an attempt to not break backwards compat.

### DIFF
--- a/src/webapp/reports/autogenerated-forms/datatables/DataTableCellSummary.tsx
+++ b/src/webapp/reports/autogenerated-forms/datatables/DataTableCellSummary.tsx
@@ -21,7 +21,9 @@ export type DataTableCellSummaryProps = {
 export const DataTableCellSummary: React.FC<DataTableCellSummaryProps> = props => {
     const { dataFormInfo, styles, dataElements, period } = props;
     const totals = _(dataElements).map(cellTotal => {
-        const values = cellTotal.items.map(itemTotal => getItemTotalDataValue(dataFormInfo, itemTotal, period));
+        const values = cellTotal.items.map(itemTotal =>
+            getItemTotalDataValue(dataFormInfo, itemTotal, period, cellTotal.strict)
+        );
         const compiled = _.template(cellTotal.formula);
         return compiled(_.merge({}, ...values));
     });
@@ -40,7 +42,8 @@ export const DataTableCellSummary: React.FC<DataTableCellSummaryProps> = props =
 export function getValueFromDataElement(
     dataFormInfo: DataFormInfo,
     itemTotal: TotalItem,
-    period?: Period
+    period?: Period,
+    strict?: boolean
 ): Maybe<number> {
     const dataElementToRead = resolveDataElement(dataFormInfo, itemTotal.dataElement);
     const dataValue = dataFormInfo.data.values.getOrEmpty(dataElementToRead, {
@@ -48,7 +51,7 @@ export function getValueFromDataElement(
         orgUnitId: dataFormInfo.orgUnitId,
         period: period || dataFormInfo.period,
     }) as DataValueNumberSingle;
-    if (dataValue.value === "") {
+    if (strict && dataValue.value === "") {
         return undefined;
     }
     const numericValue = Number(dataValue.value);
--- a/src/webapp/reports/autogenerated-forms/datatables/DataTableCellSummary.tsx
+++ b/src/webapp/reports/autogenerated-forms/datatables/DataTableCellSummary.tsx
@@ -20,13 +20,16 @@ export type DataTableCellSummaryProps = {
 
 export const DataTableCellSummary: React.FC<DataTableCellSummaryProps> = props => {
     const { dataFormInfo, styles, dataElements, period } = props;
-    const totalCalculated = _(dataElements)
-        .map(cellTotal => {
-            const values = cellTotal.items.map(itemTotal => getItemTotalDataValue(dataFormInfo, itemTotal, period));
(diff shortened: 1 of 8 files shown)


MiquelAdell (approved):

src/webapp/reports/autogenerated-forms/datatables/DataTableCellSummary.tsx
```diff
@@ -20,13 +20,18 @@ export type DataTableCellSummaryProps = {
...
+    const totals = _(dataElements).map(cellTotal => {
+        const values = cellTotal.items.map(itemTotal =>
+            getItemTotalDataValue(dataFormInfo, itemTotal, period, cellTotal.strict)
+        );
+        const compiled = _.template(cellTotal.formula);
+        return compiled(_.merge({}, ...values));
+    });
+
+    // formula evaluation in totals can return NaN if values are undefined
+    const totalCalculated = totals.every(value => Number.isNaN(Number(value)) || value === "")
```
gqcorneby, line 32: [minor suggestion] code only and not 100% sure if I got the behavior correctly. I also don't have full context on how the totals are handled in the different viewTypes so take this suggestion with a grain of salt 😅

currently, strict mode works implicitly: empty values → `undefined` → `NaN` in formulas → caught by NaN check. This obscures the intent.

Consider checking strict mode explicitly before calculating: 
[NOT TESTED]
```javascript
// Fetch all values first
const calculatedData = dataElements.map(cellTotal => {
    const values = cellTotal.items.map(itemTotal =>
        getItemTotalDataValue(dataFormInfo, itemTotal, period, cellTotal.strict)
    );
    return { cellTotal, values };
});

// Check if we should calculate based on strict mode
const shouldCalculate = calculatedData.every(({ cellTotal, values }) => {
    if (!cellTotal.strict) return true;
    
    // In strict mode, only calculate if all values are defined
    return values.every(valueRecord =>
        valueRecord !== undefined &&
        Object.values(valueRecord).every(v => v !== undefined)
    );
});

// Calculate total
const totalCalculated = shouldCalculate
    ? _(calculatedData)
          .map(({ cellTotal, values }) => {
              const compiled = _.template(cellTotal.formula);
              return compiled(_.merge({}, ...values));
          })
          .sumBy(Number)
    : "";
```


- It adds a bit of complexity but strict mode logic is visible and explicit in the calculation flow and doesn't rely on the NaN results
- Clearer intent: "in strict mode, don't calculate unless all values present"

gqcorneby (approved):
[code-only review] Approving! The description helped a lot :ok_hand: 
The added configuration makes sense for now. We can extend it to a type if there are additional empty handlings we want to specify :)

My one comment below is non-blocking - just a suggestion to make the 'strict' rule more explicit in the calculation logic.

src/webapp/reports/autogenerated-forms/datatables/DataTableCellSummary.tsx
MatiasArriola: Thanks for your feedback and digging deep into the changes, really value a fresh perspective. 

What I thought when implementing this "strict" mode is to leave up to the user formula implementation to deal with undefined values. So now, in the formula you could check for undefined, which gives you the chance to distinguish between values that are explicitly set to 0 and unset values.

In practice, the formulas we have implemented do not check for `undefined`, just operate over undefined, resulting in `NaN` values, which end up being displayed on the view as an empty string. I see this is not explicit and feels like a hack but also makes the formula more readable.

My intent was to always run the formula, while representing the unset values, which could be useful for more advanced logic in the formula. 

I think your proposed change would serve our original purpose as well, but changes the semantics a little, in that case we could also rename to something like "requiresAllValuesSet" rather than "strict"

gqcorneby: > What I thought when implementing this "strict" mode is to leave up to the user formula implementation to deal with undefined values

I see! Thanks @MatiasArriola! I misunderstood this flag. It's not about evaluation gating,
it's about value preservation (undefined vs 0)? This approach makes sense, I guess the prop "strict" is just a bit ambiguous.

Maybe `coerceEmptyToZero: false` may be clearer? Or `emptyValueMode: "zero" | "undefined"`.

Change isn't required, just a suggestion. We can also just keep it as is and document it :)

PR #144, opened 2026-02-04
[feature]: indicators headers config

MatiasArriola:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869bz6y28 #869bz6y28

### :memo: Implementation

Currently, indicators can be placed "before" or "after" a dataElement, using the `{dataset}.{section}.indicators` config. This creates a new column in the `grid` and `grid-with-cat-option-combos` viewTypes. The headers for these columns would be blank, and in this case we wanted to provide a text for an indicators column.

To solve this, we added the `{dataset}.{section}.indicatorsConfig` property. 

The existing `{dataset}.{section}.indicatorsPosition` config was moved into `{dataset}.{section}.indicatorsConfig.position` to avoid having all the indicators-related configuration scattered at the top level. 

Example config: 

```json
"indicatorsConfig": {
  "before": {
    "headers": ["literal string header"]
  },
  "after": {
    "headers": [
      {
        "code": "CONSTANT_CODE"
      }
    ]
  }
}
```

Notice there is an inconsistency carried over from previous implementation. 
"headers" was added as an array and not a single value. This is because in theory multiple indicators can be present in the same row and direction. 
However that was not accounted for when building the tables colSpans, and there is no clear way to provide an order if two or more indicators belong to the same dataElement and direction. 
In our scenario, and I think in most cases, we'll have one indicator. I left the config as an array so we could complete the implementation to support multiple indicators in the future. Also these custom headers for now are rendered into the same `CustomDataTableColumnHeader` to avoid changing the table structure.

I've noticed there is also `grid-with-periods`, but I don't see TableHead accounting for directional indicators, so I didn't add support there for this property, to avoid risks and going out of scope.


src/webapp/reports/autogenerated-forms-configurator/schemas/dataSets/section.ts
```diff
@@ -58,6 +58,28 @@ export const sectionSchema = (
         },
     };
 
+    const indicatorsConfigBeforeOrAfterSchema = {
```
gqcorneby, line 61: Nice 👏 I think updating the schema is a task that is sometimes forgotten

gqcorneby (approved):
[code-only review] everything looks good! Thanks for the detailed context in the description 👌

PR #145, opened 2026-02-04
feat: remove duplicated section.name in TableForm

MatiasArriola:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869bza8vf #869bza8vf

### :memo: Implementation

- `table` viewType was including the `section.name` as the column header. This caused it to be duplicated in relation to the the section header
- With this change the gray table header won't render, except there is a periodOffset. In that case, an empty column header was added for proper display
- Remove unused styles

We need to make sure we want this new behavior for other preexisting dataSets.

### :art: Screenshots

Before: 
imagen

After:
imagen

### DIFF
--- a/src/webapp/reports/autogenerated-forms/TableForm.tsx
+++ b/src/webapp/reports/autogenerated-forms/TableForm.tsx
@@ -11,7 +11,6 @@ import {
     DataTableRow,
     TableBody, // @ts-ignore
 } from "@dhis2/ui";
-import { makeStyles } from "@material-ui/core";
 import DataTableSection from "./DataTableSection";
 import { CustomDataTableCell, CustomDataTableColumnHeader } from "./datatables/CustomDataTables";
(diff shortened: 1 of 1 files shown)


MiquelAdell:
I'm not entirely sure that this PR introduced this issue (I'm testing in a local version of the code with all the PRs merged) but yarn threw a warning saying:
```
Line 17:8:  '_' is defined but never used  unused-imports/no-unused-imports
```

removing this line in the flagged file fixed i without breaking, it seems, anything else:
```
import _ from "lodash";
```

This line was not introduced by this PR but maybe the use of loadash was removed from this change. 
If it was please remove the import. If this is just an issue of the merging of different branches and can be ignored pelase ignore it. If it needs another way to fix that please go ahead.

Other than that I'm approving the PR

MiquelAdell (approved):

gqcorneby (approved):
Looks good, thank you @MatiasArriola!

PR #146, opened 2026-02-04
[fix]: Section toggle visibility for data elements with multiple category option combos

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869c0mm1b

### :memo: Implementation
**Problem:**
Section visibility toggles were not working correctly for data elements that have multiple category option combos (COCs). The previous implementation only checked a single data value, which caused sections to remain hidden even when data existed in other COCs.

**Solution:**
Updated the `useSectionVisibility` hook to iterate through all category option combos of a toggle data element and check if any of them have a value that satisfies the visibility condition.

### :art: Screenshots

https://github.com/user-attachments/assets/8c7b7be5-7429-4bf1-b448-075c07ad9666

#869c0mm1b

### DIFF
--- a/src/data/common/Dhis2DataFormRepository.ts
+++ b/src/data/common/Dhis2DataFormRepository.ts
@@ -793,7 +793,7 @@ function getSectionBaseWithToggle(
                     },
                 };
             } else {
-                console.warn(`Data element for toggle not found in section: ${toggle.code}`);
+                console.warn(`Data element for toggle not found: ${toggle.code}`);
                 return base;
             }
         }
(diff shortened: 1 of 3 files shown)


MatiasArriola (approved):
Code looks good to me. 
There is a small code duplication that could be extracted into an utility that gets the datavalues for all COCs of the dataElement. Maybe it could be another method in `DataValueStore` that returns `DataValue[]`.
I'm approving since it's not too important, just commenting with my point of view. 
Thanks @deeonwuli!

MiquelAdell (approved):

PR #148, opened 2026-02-05
[fix]: toggleMultiple by orgUnit with show and disabled conditions mixed

MatiasArriola:
### :pushpin: References

-   **Issue:** Closes 
     - https://app.clickup.com/t/869c1kyrm 
     - related: https://app.clickup.com/t/869c432vw (script used)

### :memo: Implementation

Some configs for `toggleMultiple` were not behaving as expected: 

```
"toggleMultiple": {
    "conditions": [
        {
            "condition": "show",
            "disabled": true,
            "orgUnits": [
                "AUT"
            ],
            "type": "orgUnit"
        },
        {
            "condition": "show",
            "orgUnits": [
                "ALB",
                "AUT"
            ],
            "type": "orgUnit"
        }
    ],
    "logicalOperator": "OR"
}
```

For example, disabled was being applied to OrgUnits outside of the rule that contained the disabled flag, but not for the ones containing disabled. 
Also disabled-only intended rules were causing hidden sections to be visible. 

This change in behavior can be better visualized testing with the current dataStore, before and after the changes.

Also, the script to modify the dataStore adding the disable rules based on CSV was added, since we've already included other TUB scripts. Updated the hide script to include the new sections.


MiquelAdell (approved):

MiquelAdell:
this has been quite tested and I'll merge it now. Please open a new PR if we need some code improvements

PR #150, opened 2026-02-10
[fix]: Section totals text to use constant codes

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869c20146

### :memo: Implementation
- [x] Fixed to properly get totals text from constants using both `texts.name` and `texts.code`
- [x] Fixed summary logic for table view types
- [x] Migrate Git hooks to Husky v5+ format: The project uses Husky v7.0.4, but the hooks were configured using the old v4 format (husky.hooks in package.json). This format is ignored in Husky v5+, meaning the pre-push checks (prettify, lint, test) were not actually running. The new husky directory format ensures:
- Tests run before push, blocking pushes with failing tests
- Code is prettified and linted before push
- Post-merge dependency checks execute properly

### :art: Screenshots

https://github.com/user-attachments/assets/b195cfb3-0c26-4821-a884-a60568377217

#869c20146


MatiasArriola (approved):
Nice one @deeonwuli!!

It is very important to have these git hooks working so we don't overlook any code issue. 

The post-merge hook was failing for me because I didn't have execution permissions on the `./.githooks/dep-check` file. I tested it directly running `./.husky/.post-merge`

Maybe you can do a `chmod +x ./.githooks/dep-check` and commit+push it. 

Anyway I'm approving since it is a detail in relation to the other changes, and not sure if it is only an issue for me. Thanks!

MiquelAdell (approved):

MiquelAdell:
I'll merge this, please create a new pr if we need some code iprovements

PR #152, opened 2026-02-13
[feature]: add section rules config to display a message for specific org units

MatiasArriola:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869c432vw

### :memo: Implementation

Important: depends on #148 since it was based on that branch. Needed base disable script, and also the toggleMultiple working as expected, although commit 07cb9f4e0da1cb8b64edf876ec6adb3758c6cae1 could be extracted to not depend on that PR.

We needed to display a message above a section for some org units. In this case, if the section is disabled, we wanted to provide a message explaining the reason.  Check the task notes for more context. 

The new config is like: 

```json
{
  "SECTION_CODE": {
    "rules": [{
      "action": {
        "text": {
          "code": "TUB_SECTION_MSG_ECDC_6"
        },
        "type": "showMessage"
      },
      "conditions": {
        "orgUnitIn": ["AUT"]
      }
    }]
  }
}
```

Some styles were provided by default, coloring resembles a warning. Maybe more generic styling can be provided, or message variants. The message has the `section-message` css class, so it could be customized via the `dataSet.customCss` option. Also the constant value will be injected as Html like happens in other parts of the app. 

I thought making it a generic placeholder "rules" with actions and conditions would be a good idea to match what we did with the `dataset.rules` config, but this could be evaluated.  
Also we might review the correct place to do the rules evaluation and mapping.


MiquelAdell:
Hi @MatiasArriola we have conflicts against 1.4.0 can you look at them? thanks!
EDIT: Oh, I see that you already mentioned that in the task and that you are off already.

@deeonwuli would you please try to fix the conflict and also review this PR? thanks!

src/data/common/Dhis2DataStoreDataForm.ts
```diff
@@ -1025,6 +1039,18 @@ export class Dhis2DataStoreDataForm {
...
+            .values()
+            .flatMap(dataSet => _.values(dataSet.sections))
+            .flatMap(section => section.rules)
+            .compact()
+            .flatMap(rule => {
+                const messageText = rule.action.text;
+                return typeof messageText !== "string" ? messageText.code : undefined;
+            })
+            .compact()
+            .value();
```
deeonwuli, line 852: This has multiple `flatMap` and `compact` calls. We could refactor this to put all the code into one `flatMap` like so:
```
 const sectionRulesCodes = _(storeConfig.dataSets)
            .values()
            .flatMap(dataSet => {
                return _(dataSet.sections)
                    .values()
                    .flatMap(section => {
                        const rules = _(section.rules)
                            .map(rule => {
                                const messageText = rule.action.text;
                                return typeof messageText !== "string" ? messageText.code : undefined;
                            })
                            .compact()
                            .value();

                        return rules;
                    })
                    .value();
            })
            .value();
```
Or extract the code to a function.

src/domain/common/entities/DataForm.ts
```diff
@@ -122,6 +123,7 @@ export interface SectionBase {
     fixedRowNames: boolean;
     hidden?: boolean;
     columnsConfig?: Record<string, { rules?: RulesFormula }>;
+    rules?: SectionRule[];
```
deeonwuli: [nitpick]: Why do we opt to use the optional property here? If the rule is not defined, we could return an empty string instead

src/webapp/reports/autogenerated-forms/hooks/useSectionRuleActions.ts
```diff
@@ -0,0 +1,25 @@
...
+    sectionId: string;
+    dataFormInfo: DataFormInfo;
+}): UseSectionRuleActionsResult {
+    return React.useMemo<UseSectionRuleActionsResult>(() => {
+        const section = dataFormInfo.metadata.dataForm.sections.find(s => s.id === sectionId);
+        const messages = section?.rules
+            ?.filter(rule => rule.action.type === "showMessage")
+            .map(rule => rule.action.text);
+        return {
+            messages: messages,
```
deeonwuli, line 20: We could return an empty array here when `messages` or `section` are undefined

src/domain/common/entities/SectionRule.ts
```diff
@@ -0,0 +1,33 @@
...
+};
+
+function evaluateSectionRuleCondition(rule: SectionRule, context: SectionRuleContext): boolean {
+    if (rule.conditions.orgUnitIn) {
+        return rule.conditions.orgUnitIn.includes(context.orgUnitCode);
+    }
+    return false;
+}
+
+export function getApplicableSectionRules(rules: SectionRule[] | undefined, context: SectionRuleContext): SectionRule[] {
```
deeonwuli: `rules` do not have to be `Maybe<SectionRule[]` if section rules default to an [] when they are undefined.

src/scripts/build-tub-ou-toggle-disable.ts
```diff
@@ -0,0 +1,283 @@
...
+    logicalOperator: "OR" | "AND";
+    conditions: OrgUnitToggleCondition[];
+};
+
+type SectionConfig = {
+    toggleMultiple?: ToggleMultiple;
+    rules?: SectionRule[];
+};
+
+type Config = Record<"dataSets", Record<Code, { sections: Record<Code, SectionConfig> }>>;
```
deeonwuli, line 27: We could move these types to a shared types folder for the ou toggle scripts. These exact types are used in 3 places.

src/scripts/build-tub-ou-toggle-disable.ts
```diff
@@ -0,0 +1,283 @@
...
+    console.debug(`Output written to ${outputFilePath}`);
+}
+
+main();
+
+async function parseConfigFromFile(configPath: string): Promise<Config> {
+    const text = await readFile(configPath, "utf8");
+    const autogenConfig: Config = JSON.parse(text);
+
+    return autogenConfig;
```
deeonwuli, line 136: This can also be moved to the shared file. It is repeated in several files

src/scripts/build-tub-ou-toggle-disable.ts
```diff
@@ -0,0 +1,283 @@
...
+            },
+        };
+        Object.values(updatedConfig.dataSets).forEach(dataSetConfig => {
+            updateSectionConfigWithMessages(dataSetConfig, section, ruleObj);
+        });
+    });
+    return updatedConfig;
+}
+
+function parseRulesFromExcel(rulesSheetPath: string): CsvRow[] {
```
deeonwuli, line 272: This is repeated as well

deeonwuli (changes requested):
Like you suggested, we would definitely benefit from a refactor for the rules evaluation and mapping as rules are evaluated in the domain layer, in a data repository and in hooks. 

Code is good. Just some minor suggestions. 👍

MiquelAdell (approved):

src/domain/common/entities/DataForm.ts
MatiasArriola: I've used optional property there because to me it looked semantically correct, other properties were already optional, and also thought it could be a good idea for backwards compatibility and safety.

But thanks to your comment I've realized I was already always populating the property at the repo level with an empty array if there are no rules in the config, so updated it to be required

src/data/common/Dhis2DataStoreDataForm.ts
MatiasArriola: I've made a small refactor there to reduce the compact calls and also extracted a small `getCode` util to avoid repeating code. Just let me know if you want to further discuss this point

MatiasArriola:
Thanks for the review @deeonwuli.

Sorry I didn't mention in the PR description, but I think the tub-related scripts are temporary, and all of them should be deleted: 
```
"build-tub-ou-toggle-hide": "npx ts-node -P tsconfig-cli.json src/scripts/build-tub-ou-toggle-hide.ts",
"build-tub-ou-toggle-disable": "npx ts-node -P tsconfig-cli.json src/scripts/build-tub-ou-toggle-disable.ts",
"build-tub-ou-toggle": "npx ts-node -P tsconfig-cli.json src/scripts/build-tub-ou-toggle.ts", 
```
They do not belong to the d2-autogen-forms repo because they are specific to a particular project instance, and only needed for the initial setup. That's why I haven't invested any time on following best practices there, focusing on implementation speed. Only wanted to keep a copy somewhere in case it was needed. 

- If we decide to keep these scripts anyway, I must apply a refactor. 
- Otherwise, I could create a separate PR to delete them, and keep them in this branch in case we need to look back in the future.

MiquelAdell:
Hi @MatiasArriola,
yes, tub scripts are temporary and should not be here.
Can you please move them to this repo instead?
https://github.com/EyeSeeTea/TUB_scripts

src/data/common/Dhis2DataStoreDataForm.ts
deeonwuli: Nice 👌

deeonwuli (approved):
All good. Thanks @MatiasArriola

MiquelAdell:
@MatiasArriola I made some merges in development and now we have conflicts. Could you look to them and then we can merge this PR?
Thanks!

PR #160, opened 2026-03-13
feat(tabs): add bottom navigation bar for sequential tab traversal

ifoche:
## Summary

- Add opt-in bottom navigation bar (`Previous` / `X/Y` indicator / `Next`) for sequential tab traversal in autogenerated forms
- Controlled by `showNavigation: true` in the dataset's DataStore config
- Navigation counts only visible tabs (excludes hidden tabs and the "Others" tab)

## DataStore Config

Add to the dataset entry in `d2-reports/autogenerated-forms`:

```json
{
  "dataSets": {
    "<dataSetId>": {
      "showNavigation": true
    }
  }
}
```

## Changes

- **Domain**: Added `showNavigation: boolean` field to `DataForm` entity
- **Data layer**: Extended `Dhis2DataStoreDataForm` codec and mapped config to entity
- **Presentation**: New `TabNavigation` component, integrated into `SectionsTabs`
- **Tests**: Unit tests for codec parsing (4 tests) and TabNavigation component (14 tests)

## Related Tasks
- [Add next/prev button in autogen forms](https://app.clickup.com/t/869ceydey)

## Test Plan
- [x] Enable `showNavigation: true` for a dataset with multiple tabbed sections
- [x] Verify Previous is disabled on the first tab
- [x] Verify Next is disabled on the last tab
- [x] Click Next and Previous to navigate between tabs
- [x] Verify X/Y indicator updates correctly
- [x] Hide some tabs via visibility rules and verify hidden tabs are excluded from count
- [x] Verify "Others" tab is excluded from navigation
- [x] Remove `showNavigation` from config and verify navigation bar disappears


MiquelAdell:
1. I get a warning on compile:
```
Compiled with warnings.

src/webapp/reports/autogenerated-forms/SectionsTabs.tsx
  Line 317:26:  Forbidden non-null assertion  @typescript-eslint/no-non-null-assertion
```

MiquelAdell:
About the "others" tab. This is in the development branch and not merged against 1.4.0
We'll need to check that once 1.4.0 is merged into development
Meanwhile I'll ask to merge development into 1.4.0 and we can check if this PR is still open but, if not, it can be closed and we can open a new one if it generates any issue

MiquelAdell (approved):

MiquelAdell:
Hi @MatiasArriola can you please do a code review of this functionality?

src/webapp/reports/autogenerated-forms/SectionsTabs.tsx
```diff
@@ -246,6 +247,38 @@ const SectionsTabs: React.FC<TabPanelProps> = React.memo(props => {
...
+                            : undefined;
+                    if (value !== visibleRule.formula.condition) return [];
+                }
+
+                const label = section.texts.tabLabel || section.name;
+                return [{ primaryIndex, label }];
+            }),
+            tab => tab.primaryIndex
+        );
+    }, [sections, tabVisibilityByIndex, dataFormInfo]);
```
MatiasArriola: This should belong to the hook `useTabVisibility` and reuse logic from there.
Returning the memoized `visiblePrimaryTabs` from that custom hook would be appropriate, reusing `tabVisibilityByIndex` / `sortedVisibleTabIndices`.

Not just a thing of code organization, but we must aim to avoid duplicating tab visibility logic.

src/webapp/reports/autogenerated-forms/SectionsTabs.tsx
```diff
@@ -277,6 +310,14 @@ const SectionsTabs: React.FC<TabPanelProps> = React.memo(props => {
         }
     }, [activeTab, tabVisibilityByIndex, firstVisibleTabIndex]);
 
+    useEffect(() => {
+        if (visiblePrimaryTabs.length === 0) return;
+        const isActiveTabVisible = visiblePrimaryTabs.some(tab => tab.primaryIndex === activeTab);
+        if (!isActiveTabVisible) {
+            setActiveTab(visiblePrimaryTabs[0]!.primaryIndex);
+        }
+    }, [activeTab, visiblePrimaryTabs]);
```
MatiasArriola: This new `useEffect` duplicates the intent of the previous `useEffect`: 

    useEffect(() => {
        if (firstVisibleTabIndex === undefined) return;
        if (!tabVisibilityByIndex[activeTab]) {
            setActiveTab(firstVisibleTabIndex);
        }
    }, [activeTab, tabVisibilityByIndex, firstVisibleTabIndex]);

Which is setting the active tab to the first visible tab.

MatiasArriola (changes requested):
See inline comments for requested changes. 

Also, make sure to fix any issue in `yarn lint` and type issues  `yarn tsc --noEmit`. There were some type issues that might be unrelated to this PR.

The feature works well for me in the TUB dataset. 

Not code related: it might be nice to automatically scroll to the top upon changing tabs. Not sure, just a thing to consider. 

Thanks!

MiquelAdell:
We have made changes to development. They will need to be addressed before the merge

MiquelAdell:
I agree with Matias on the "scroll to top on tab change"

ifoche:
Hi @MatiasArriola @MiquelAdell — all review feedback has been addressed:

**Code changes (commit d6e6fbf):**
- Moved `visiblePrimaryTabs` computation into `useTabVisibility` hook, reusing existing `tabVisibilityByIndex`/`sortedVisibleTabIndices` — no more duplicated tab visibility logic
- Extracted `isFormulaVisible` helper in the hook to consolidate formula-based visibility checks
- Removed the duplicate `useEffect` for active tab reset (the original one already handled it)
- Fixed the lint warning (non-null assertion)
- Fixed `yarn tsc --noEmit` errors in `orgUnitFixtures.ts` (missing `code` property)

**Merge + scroll-to-top (commit 01089c7):**
- Merged `development` into the feature branch and resolved conflicts (SectionsTabs, useTabVisibility, Dhis2DataStoreDataForm, orgUnitFixtures)
- Added `showNavigation` to the new `AutogenConfig.ts` `DataSetConfig` type
- Added scroll-to-top (`window.scrollTo(0, 0)`) on **all** tab changes — both top tab bar clicks and bottom navigation bar (Previous/Next)

All checks pass: `tsc` (0 errors), `lint` (0 warnings), `test` (61/61 passing).

MiquelAdell:
I created a new section and did not assign it to any tab and now I get an "others" tab (this is related to another PR and it works as expected).
But the paging is not counting this "other" section.
see the screenshots as an example:
image
image
In this example we have 18 sections plus "other" so 19 sections 
See that, in the pager, we are counting 18/18 and next 18/0
We should have 18/19 and then 19/19

Note: do not worry about the numbering on the tabs above, this is correct as it is

ifoche:
@MiquelAdell Fixed in commit 5fd19e0 — the "Others" tab is now included in the navigation pager count. It should show e.g. 18/19 and 19/19 correctly, and Previous/Next will navigate to/from the Others tab as well. Ready for re-test!

MiquelAdell:
> @MiquelAdell Fixed in commit [5fd19e0](https://github.com/EyeSeeTea/d2-autogen-forms/commit/5fd19e0093f5fc7110577c8891c94a885376f8ae) — the "Others" tab is now included in the navigation pager count. It should show e.g. 18/19 and 19/19 correctly, and Previous/Next will navigate to/from the Others tab as well. Ready for re-test!

works correctly. Thanks.
Just waiting the final code review from @MatiasArriola

src/webapp/reports/autogenerated-forms/hooks/useTabVisibility.ts
```diff
@@ -41,5 +51,45 @@ export function useTabVisibility(props: TabPanelProps): TabVisibilityState {
...
+            tab => tab.primaryIndex
+        );
+
+        const othersTab: VisibleTab[] = tabVisibilityByIndex[-1] ? [{ primaryIndex: -1, label: "Others" }] : [];
+
+        return [...tabbedVisible, ...othersTab];
+    }, [tabbedSections, tabVisibilityByIndex, dataFormInfo]);
+
+    return { tabVisibilityByIndex, firstVisibleTabIndex, visiblePrimaryTabs };
+}
```
MatiasArriola, line 84: `visiblePrimaryTabs` duplicates logic that's already in `tabVisibilityByIndex` (index parsing, visibility checks) and adds an extra `isFormulaVisible` check that `tabVisibilityByIndex` doesn't know about, making them inconsistent on what "visible" means.

This same formula check already exists in `SectionsTabs.tsx` (lines 354-368), so we now have several places computing the same thing. 

I think we can add the `isFormulaVisible` check into the `tabVisibilityByIndex` computation, so it becomes the single source of truth for tab visibility. Then derive `visiblePrimaryTabs` from `tabVisibilityByIndex` + labels (no re-checking visibility, use "Others" label for -1), and have `SectionsTabs.tsx` rely on `tabVisibilityByIndex` too instead of recomputing the formula inline.

MatiasArriola (changes requested):
Looking good. I left one comment, please let me know if that does not make sense.

ifoche:
@MatiasArriola Addressed your latest review in d11a9e3:

- **`tabVisibilityByIndex` is now the single source of truth** — it incorporates both section visibility (`getSectionVisibilityState`) and formula visibility (`isFormulaVisible`) in one computation. A tab is visible only if it has at least one visible section AND its header's formula passes.
- **`visiblePrimaryTabs` just derives from `tabVisibilityByIndex`** — no re-checking formula, no duplicated logic.
- **Removed inline formula computation from `SectionsTabs.tsx`** — the `visibleRule`/`calculateFormula`/`isConditionMet` block is gone; the component now relies solely on `tabVisibilityByIndex`.

Let me know if this aligns with what you had in mind!

MatiasArriola (changes requested):
I ran another round of testing and review of this PR. Some updates: 

- When some tabs are hidden, the bottom navigation only counts visible tabs. i.e. if tab 12 is hidden and you go to tab 13, bottom nav will show you are in tab 12. 
  - [test url](http://localhost:8081/?orgUnit=oofyLUJJ6Vy&dataSet=XGn6mrO2HU8&period=2025) for local.
  - maybe current behavior is ok, just found it a little bit misleading (header saying I'm on 16, bottom bar saying 14). Also to review the total part. i.e. show 21 (last section is 20 + others) instead of 19.
- Regarding code quality: 
  - the `visiblePrimaryTabs` memo part still contains some duplicated logic. i.e. it does ` const primaryValue = _(section.tabs.order).split(".").first() ?? "0";`, while it could reuse `const [primaryTabIndex] = getTabIndices(section.tabs.order);` 
  - missed this from previous feedback, but maybe worth a look: a new `VisibleTab` type was introduced. It has the `label` property. Most of the code in that `visiblePrimaryTabs` block is to build this object. But the `label` property is never used. So we are introducing unnecessary complexity. 

EDIT: 
I see the tab index/total when hidden tabs exist is working according to the specs included alongside the code: 
```
#### Scenario: Some tabs hidden by rules
- **WHEN** there are 5 primary tabs but tabs 2 and 4 are hidden by visibility rules, and the user is on the originally-third tab
- **THEN** the indicator shows "2/3" (only visible tabs 1, 3, 5 are counted)
```

MiquelAdell:
Regarding this comment from @MatiasArriola 
> When some tabs are hidden, the bottom navigation only counts visible tabs. i.e. if tab 12 is hidden and you go to tab 13, bottom nav will show you are in tab 12.
[test url](http://localhost:8081/?orgUnit=oofyLUJJ6Vy&dataSet=XGn6mrO2HU8&period=2025) for local.
maybe current behavior is ok, just found it a little bit misleading (header saying I'm on 16, bottom bar saying 14). Also to review the total part. i.e. show 21 (last section is 20 + others) instead of 19.

I agree it is confusing but I don't have a good solution for this. The client requested to keep the numbering so if we are hiding section 15 section 16 is labeled 16, not 15.
So we end up with skipped numbers in the section numbers and a total number of sections at the end that may be also different. 
But I think we can keep it like this. 

regarding this:
> the visiblePrimaryTabs memo part still contains some duplicated logic. i.e. it does  const primaryValue = _(section.tabs.order).split(".").first() ?? "0";, while it could reuse const [primaryTabIndex] = getTabIndices(section.tabs.order);

Do you mean that you already flagged this and it was not addressed? If that's the case, I want to know because I don't want to be running in circles and having something not being addressed. 

> missed this from previous feedback, but maybe worth a look: a new VisibleTab type was introduced. It has the label property. Most of the code in that visiblePrimaryTabs block is to build this object. But the label property is never used. So we are introducing unnecessary complexity.

Sure, let's fix that one by removing this label that we do not need 

Also, please, claude, merge development into this branch and fix conflicts

MatiasArriola:
> I agree it is confusing but I don't have a good solution for this. The client requested to keep the numbering so if we are hiding section 15 section 16 is labeled 16, not 15. So we end up with skipped numbers in the section numbers and a total number of sections at the end that may be also different. But I think we can keep it like this.

OK!

> Do you mean that you already flagged this and it was not addressed? If that's the case, I want to know because I don't want to be running in circles and having something not being addressed.

Well, mentioned "index parsing" in my previous comment, but to be fair the `getTabIndices` function does a bunch of other things. At least I would try to avoid doing the `split('.')[0]` in all places, there should be a central place to get the tab number from.

Thanks!

MatiasArriola (approved):
Thanks for the updates!

In my previous comment when I wanted to remove the `label` property that was not used in `VisibleTab`, I would also have removed the `VisibleTab` type entirely and keep track of visible tabs as an array of numbers, but I did not mention it explicitly (I might be nitpicking here). 

Since he change is working and it is well scoped, I'm approving
Thanks!

PR #163, opened 2026-03-17
[feature]: handle nbsp in separators

MatiasArriola:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869ce25c7 #869ce25c7

### :memo: Implementation

- Separator logic was spread in all viewModels. Refactored and extracted into `src/webapp/reports/autogenerated-forms/utils/dataElementNameSeparator.ts`
- Most of the times, we were splitting dataElement names by ` - `. This became an issue with some translations. For example, in french, non-breaking space characters (`&nbsp;` or `\u00A0`) were commonly used, causing the different views to not group/render correctly.
- These changes are mostly a refactor, adding `\u00A0` into consideration for the separators, and minor optimizations. No other behavior change should be there.


MiquelAdell:
@gqcorneby a quick code review would be enough for this PR.

@MatiasArriola The form does not break at all but I'm not sure if you removed the code that transformed non-break-white-space (nbws) to white-space (ws).

I have reimported the translations with the nbws but I can not see the nbws in the form afterwards.
I imported this file
(see line 147)
Screenshot 2026-03-19 at 16 45 20
and when I print the form and I copy and paste this line into VScode:
image
I get this:
Screenshot 2026-03-19 at 16 45 24

It seems we have ws before the ":"
could it be that we are still replacing it?

I think we introduced the "fix", replacing nbws by ws in another PR but then maybe this PR should remove that?

thanks!

MatiasArriola:
You are right @MiquelAdell, I missed that. With this PR, the name parts are joined again using common whitespaces, and the nbsp is not kept (`joinDataElementName`). This was not intentional, but mostly to stay in line with how it was working previously (split names then join). I handled the nbsp at split level, but at the join level it is still using only the ` - ` separator. 

If we need to keep the nbsps intact when displaying them in the form, I could revise this PR. It would add some complexity or might require a larger refactor, because we would need to keep track of the original separator match.

MiquelAdell:
oh, I thought it was a "issue" not a concious decission. Let's leave it like it is right now as we do not have time in the project.

Thanks!

MiquelAdell (approved):

gqcorneby (approved):
Thanks @MatiasArriola! Now handling the patterns are centralized 👌 Everything looks good.

PR #164, opened 2026-03-19
feat(comments): visual indicator for data values with comments

ifoche:
## Summary
- Adds `comment` field to `DataValueBase` and extracts it from the DHIS2 API response (already returned but previously discarded)
- `CommentIcon` now shows a small green dot badge when the data value has an existing comment
- Wired through `DataElementItem` — all form types (table, grid, grid-with-periods, etc.) inherit the indicator automatically
- Enabled by default for all datasets, no configuration needed
- TDD approach: 8 new tests in `CommentIndicator.spec.ts`

## Related Tasks
- [Estimate comments-related development](https://app.clickup.com/t/869chvhvy)
  - [Domain: Add comment field + tests](https://app.clickup.com/t/869cjedm2)
  - [Data: Extract comment from API](https://app.clickup.com/t/869cjedmd)
  - [Presentation: CommentIcon badge](https://app.clickup.com/t/869cjedn2)
  - [Wiring: DataElementItem → CommentIcon](https://app.clickup.com/t/869cjednj)
  - [Validation: Visual + regression](https://app.clickup.com/t/869cjedp0)

## Test plan
- [x] Verify green dot badge appears on comment icons for data values with existing comments
- [x] Verify no badge on data values without comments
- [ ] Test across form types: table, grid, grid-with-periods
- [x] Verify clicking the comment icon still opens the DHIS2 comment dialog
- [x] Run `yarn test` — all 47 tests pass
- [x] Run `yarn tsc --noEmit` — no type errors
- [x] Run `yarn lint` — clean


src/domain/common/entities/DataValue.ts
```diff
@@ -17,6 +17,7 @@ export interface DataValueBase {
     period: Period;
     categoryOptionComboId: Id;
     isRequired?: boolean;
+    comment?: string;
```
eperedo: I think making `comment` an optional parameter could lead to unexpected errors. In this case we're not assigning the comment property in the method `checkCompulsoryAndBuildDataValues` in file `Dhis2DataValueRepository`. 

Also it protect us if we `forgot` to pass the comment any time we need to use the dataValue entity. We can add the fallback to an empty string as a default value.

src/domain/common/entities/DataValue.ts
```diff
@@ -204,4 +205,8 @@ export function getEmpty(dataElement: DataElement, base: DataValueBase): DataVal
     }
 }
 
+export function hasComment(dataValue: Pick<DataValueBase, "comment">): boolean {
+    return !!dataValue.comment;
```
eperedo: Subjective, but I prefer `dataValue.comment.length > 0` or `Boolean(dataValue.comment)`.

src/webapp/reports/autogenerated-forms/DataElementItem.tsx
```diff
@@ -47,6 +47,15 @@ export const DataElementItem: React.FC<DataElementItemProps> = React.memo(props
     const elId = _([dataElement.id, period]).compact().join("-");
     const dataElementCocId = dataElement.cocId ?? dataFormInfo.categoryOptionComboId;
     const auditId = _([dataElement.orgUnit, dataElement.id, dataElementCocId, "val"]).compact().join("-");
+    const hasComment =
+        !noComment &&
+        dataValueHasComment(
+            dataFormInfo.data.values.getOrEmpty(dataElement, {
+                orgUnitId: dataFormInfo.orgUnitId,
```
eperedo: For `orgUnitId` it should be `dataElement.orgUnit || dataFormInfo.orgUnitId`. It will cover the case for the subnational dataSet where we need to save dataValues in a different orgUnit than the selected in the dataEntry.

Also using `React.memo` for the `hasComment` variable would be good.

eperedo (changes requested):

ifoche:
@eperedo Thanks for the review! All three points have been addressed in e3aa009:

1. **`comment` is now required** (`string`, default `""`) on `DataValueBase` — added `comment: ""` to `Dhis2DataValueRepository.checkCompulsoryAndBuildDataValues` and all test fixtures. Introduced a `DataValueLookup` type for `get`/`getOrEmpty` so existing callers (which only need the key fields) don't have to pass `comment`.

2. **`hasComment`** now uses `dataValue.comment.length > 0` instead of `!!`.

3. **`DataElementItem.tsx`** — `orgUnitId` now falls back correctly with `dataElement.orgUnit || dataFormInfo.orgUnitId` for subnational datasets, and the `hasComment` computation is wrapped in `React.useMemo`.

src/domain/common/entities/DataValue.ts
```diff
@@ -108,6 +108,9 @@ export type DataValue =
     | DataValueDate
     | DataValueMultiText;
 
+export type DataValueLookup = Pick<DataValueBase, "orgUnitId" | "period" | "categoryOptionComboId"> &
+    Partial<Pick<DataValueBase, "comment" | "isRequired">>;
```
eperedo: The new type `DataValueLookup`  is correct, but we don't need the `comment` or `isRequired` in the methods: `get`, `getOrEmpty` or `getEmpty`.

I think `export type DataValueLookup = Pick<DataValueBase, "orgUnitId" | "period" | "categoryOptionComboId">` is enough, right?

src/webapp/reports/autogenerated-forms/MirrorDataElementItem.tsx
```diff
@@ -21,7 +21,7 @@ export function getMirrorDataValue(
     base: Pick<DataValueBase, "period" | "categoryOptionComboId" | "isRequired">
 ): DataValue {
     const resolvedDataElement = resolveDataElement(dataFormInfo, dataElement);
-    const dataValueBase: DataValueBase = {
+    const dataValueBase = {
```
eperedo: And here we can use `const dataValueBase: DataValueLookup = { ... }` instead removing the type. Just to be extra safe in case we modify the object in the future with the type TS will yell at us if we introduce a wrong property.

eperedo (changes requested):
Looks good, just a couple of things related to TS.

src/domain/common/entities/DataValue.ts
MiquelAdell: Addressed in 708d686 — `DataValueLookup` now just picks `orgUnitId | period | categoryOptionComboId`. Callers already covered `comment`/`isRequired` via the default in `getEmpty` and optional typing.

src/webapp/reports/autogenerated-forms/MirrorDataElementItem.tsx
MiquelAdell: Addressed in 708d686 — `dataValueBase` is now typed `DataValueLookup`. Also dropped the unused `isRequired` passthrough since no caller sets it and the lookup type no longer carries it.

MiquelAdell (approved):
the functionality works but it loads comment information on form load. Making it refresh after you leave a comment would be great but it can be done in a future PR

eperedo (approved):

PR #170, opened 2026-03-26
[feature]: Give dataElement toggle conditions precedence over orgUnit in toggleMultiple

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869ckx3p6

### :memo: Implementation 
**Problem**:
Previously, all toggle conditions (dataElement and orgUnit) were evaluated together in a sinfgle list. This meant orgUnit conditions could override dataElement conditions depending on the logical operator — e.g., with OR, truthy orgUnit conditions could make a section hidden even when a dataElement condition explicitly said it should be visible.

**Solution**:
DataElement conditions are evaluated first; if they produce a definitive result (true/false), it is used. OrgUnit conditions are only used as a fallback

### :art: Screenshots

https://github.com/user-attachments/assets/bfbfa68d-fff2-4318-a2fb-b8220da42632

#869ckx3p6


MiquelAdell (approved):

src/webapp/reports/autogenerated-forms/hooks/useSectionVisibility.ts
```diff
@@ -89,41 +91,93 @@ type ToggleMultipleConditionState = {
...
+        _(toggleDataElements)
+            .map(toggle => {
+                if (toggle.type !== toggleType) return undefined;
+                return {
+                    isVisible: evaluateToggleCondition(toggle, dataFormInfo, currentOrgUnitCode),
+                    isDisabled: toggle.dataElement.disabled,
+                };
+            })
+            .compact()
+            .value();
```
MatiasArriola: [Minor / code style] Instead of mapping -> return undefined in some cases -> compact, we could make the intent more explicit with something like this:

```typescript
const getToggleConditions = (toggleType: "dataElement" | "orgUnit") =>
        toggleDataElements
            .filter(toggle => toggle.type === toggleType)
            .map(toggle => {
                return {
                    isVisible: evaluateToggleCondition(toggle, dataFormInfo, currentOrgUnitCode),
                    isDisabled: toggle.dataElement.disabled,
                };
            });
```

Also, even though I know we extensively use lodash in our projects, for example here we would be covered with native array functions.

src/webapp/reports/autogenerated-forms/hooks/useSectionVisibility.ts
```diff
@@ -89,41 +91,93 @@ type ToggleMultipleConditionState = {
...
-            : conditions.some(condition => condition.isVisible);
-
-    const visibleConditions = conditions.filter(condition => condition.isVisible);
-    const isDisabled =
-        logicalOperator === "AND"
-            ? visibleConditions.every(condition => condition.isDisabled)
-            : visibleConditions.some(condition => condition.isDisabled);
-    return { isVisible, isDisabled };
+    return logicalOperator === "AND"
+        ? _(conditions).every(condition => evalFunction(condition))
```
MatiasArriola: Just a minor detail, but similar comment about lodash usage here. We would be fine with doing `conditions.every` without the need of the lodash wrapper

MatiasArriola (approved):
Thanks @deeonwuli, looks good!

If you agree, you could implement the minor changes from my comments. But they are only non-blocking details so I'm approving.

deeonwuli:
> Thanks @deeonwuli, looks good!
> 
> If you agree, you could implement the minor changes from my comments. But they are only non-blocking details so I'm approving.

I agree. Using native array functions completely removes the need for lodash in the file.

PR #176, opened 2026-04-10
fix: sort DisaggregatedCOCsGrid columns/rows by metadata order

MiquelAdell:
The previous sort logic ordered the table axes by the (translated) display label, with a hard-coded "unknown"/"other" pattern to push those to the end. That broke for non-English translations: in Spanish "desconocido" sorts before "resistente" / "susceptible" alphabetically and the unknown-pattern matcher does not catch it, so the column order changes per language.

Sort instead by the canonical position of the category option in its parent DHIS2 Category. The d2-api schema does not expose a sortOrder field on CategoryOption, so derive it from the array index of the categoryOptions collection on the parent Category — DHIS2 returns this collection in its canonical sort order set in the maintenance app. The index is propagated through getCocOrdered and getVisibleCategoryOptionCombos via a per-data- element lookup map, and exposed on the CategoryOption domain entity as a new `sortOrder: number` field.

The hard-coded unknown-pattern bucket and numeric-prefix parsing in sortItems are deleted; the function is now a one-line sortBy on the new field. This resolves the TODO at the top of sortItems about moving the patterns into datastore config — they are no longer needed.

### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869cv518f

### :art: Screenshots

before (in Spanish)
Screenshot 2026-04-10 at 09 47 29
after
Screenshot 2026-04-10 at 09 48 41

the code is generated by claude


MatiasArriola (approved):
Thanks @MiquelAdell!

The PR looks good and is a good improvement over the existing sort. I didn't test, only looked at the code myself and also ran reviews with LLMs. 

I'm approving, but some comments worth checking:

- `Dhis2DataElement.ts#buildSortOrderByCode`: The function flattens all categories into a single code → index map. If two different categories share a category option with the same code, the second one silently overwrites the first.
  - This might not be true (or work fine in practice) but worth commenting.
-  `Dhis2DataElement.ts:264-279`:  line 278 assigns `sortOrder: index` directly from the array position. Meanwhile, `categoryOptionCombos` paths use `sortOrderByCode[co.code]` which is built from the same array indices. These two paths should produce the same values, but `buildSortOrderByCode` could lose data on code collisions (previous comment), while the direct index cannot. This inconsistency could lead to columns and rows sorting differently for the same category option.
- Deleted `getCocFormName` function changed behavior subtly                                                                                                                                           
  - The old `getCocFormNameFromCombos` returned a plain string via `getCocFormName`, which fell back to `cocName` (the first part of the split). The new version falls back to `optionCode` instead. Verify this is intentional and possible impact.

MiquelAdell:
I don't think any of the three items need work.

The first is a non-issue in practice because categories in a disaggregated grid don't share options so that makes 1 and 2 not need any work.

For 3 
Before we had:
```
getCocFormNameFromCombos → getCocFormName → returns displayFormName || cocName
                                            where cocName = the second part of COC name split by ", "
```
The old fallback when no displayFormName was found was cocName — which was actually the optionCode parameter passed in (confusingly named cocName in the old getCocFormName). So the variable was called cocName but it held the category option code from getRowCategoryOptionCodeFromCoc.

Now we have:
```
getCocFormNameFromCombos → returns matchedOption?.displayFormName || optionCode
```
The fallback is now explicitly optionCode — which is the exact same value that flowed through as cocName in the old code. Look at the call site: getCocFormNameFromCombos(cocs, rowCatOptionCode) — the second argument is rowCatOptionCode, which comes from splitting the COC name by ", " and taking index [1]. That value was passed as cocName to getCocFormName, and now it's called optionCode in the new version.

The behavior is identical. The old code just had a misleading variable name (cocName for something that was actually a category option code). The refactor made the naming honest. No functional change in the fallback.

So I'm merging

thanks!

PR #177, opened 2026-04-10
[fix]: Apply rules to sections with period offset

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869cmxcc8

### :memo: Implementation
Fixes rule evaluation for sections that use a period offset. Previously, the getValueAndVerifyCondition function used the period parameter (which could be an offset-adjusted period from a section) when looking up data values for rule conditions. This caused rules to evaluate against the wrong period's data, leading to incorrect rule application.

#869cmxcc8

### DIFF
--- a/src/webapp/reports/autogenerated-forms/AutogeneratedForm.tsx
+++ b/src/webapp/reports/autogenerated-forms/AutogeneratedForm.tsx
@@ -313,7 +313,13 @@ function useDataFormInfo() {
                 checkValidationRules(dataValue);
             }
         },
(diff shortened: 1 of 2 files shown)


MiquelAdell (approved):

MatiasArriola (approved):
At code level the fix looks good to me. It will use the current form period instead of the sections period which may contain offsets, while `evaluateTotalRule` still uses the `period` parameter.
 
Maybe we could add a comment as copilot suggested to clarify this behavior for future readers.

Thanks @deeonwuli!

MiquelAdell:
@deeonwuli could you please add the comment @MatiasArriola  suggested? thanks!

PR #179, opened 2026-04-20
[feature]: Inline constant creation in configurator

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869ckzqxc

### :memo: Implementation
- [x] Adds an "inline constant creation" workflow to the Autogen Configurator. While editing the dataStore JSON in the Monaco editor, when the cursor lands inside a `"code": "..."` value (any `texts.*.code` reference), the suggestion widget shows a **Create new constant** entry. Selecting it opens a dialog to define the constant (name, short name, code, description) and saves it to DHIS2.  
- [x] The change is also captured as a spec-driven proposal under `openspec/changes/inline-constant-creation/` (`proposal.md`, `design.md`, `tasks.md` + delta spec) and lives long-term as `openspec/specs/inline-constant-creation/spec.md`.   
- [ ] Add a translations field to `CreateConstantDialog` so users can upload different translations.

### :art: Screenshots

https://github.com/user-attachments/assets/ee48d6ac-2614-4467-90d6-4787c2a64011


Test against `https://dev.eyeseetea.com/who-dev-41/` with a user that has `F_CONSTANT_ADD` (or `ALL`).                                                                                                                   
1.  **Small config** (e.g. *MAL - Malaria Free*): cursor inside any `texts.*.code` value → widget shows a single `Create new constant` entry. Selecting it opens the dialog. Save creates the constant in DHIS2 and inserts its code into the JSON at the cursor.                                                                                                                                                                  
2.  **Large config** (e.g. *TUB - Annual data*, ~300KB): same as above. The "Large file detected" banner is expected, schema validation and the JSON-worker's string completions are intentionally disabled for performance.                                                                                                                                                                                                             
3.  **Permissions**: log in as a user without `F_CONSTANT_ADD`. The create entry should appear marked deprecated (struck through) with detail `Requires F_CONSTANT_ADD authority`; selecting it should be a no-op.
4.  **Validation**: leave any required field blank and click Save → the form should show per-field errors and not call the API. Submitting a `shortName` longer than 50 characters should also block.
5. **Duplicate code error**: try saving with a `code` that already exists in DHIS2. The dialog should surface a per-field error on `code`.
                                                   
#869ckzqxc


xurxodev (changes requested):
thanks @deeonwuli 

Solid feature PR — the layering is right (codec → entity helpers → use case → repo → UI), the Monaco completion is encapsulated in its own module, and the design doc + tests for the JSON-path walker make the trickiest piece reviewable. Most comments are about regressions in the shared `ConstantRepository.save` contract, error visibility, and a few `any`s that should be typed.

## 1. Should fix

- **`src/data/ConstantD2Repository.ts:15-50` — chunking regression**. The previous `save` chunked `ids` in groups of 100 and POSTed each chunk independently. The new `buildConstantPayload` collapses everything into a single POST (existing items chunked only for the *fetch* step, but then concatenated and sent in one shot). For the inline-create path (one item) this is fine; for `ImportConstantTranslationsUseCase` and any future bulk import it's a regression. Either restore chunking on the POST, or assert that `save` is only used for small batches and update the docstring of the repository method.
- **`src/data/ConstantD2Repository.ts` — empty-array behaviour**. With the rewrite, `save([])` now POSTs `{ constants: [] }` (before, the chunk loop made it a no-op). Add an early `return Stats.empty()` when both `existingConstants` and `newConstants` are empty.
- **`src/domain/common/entities/User.ts:14` — stray comment**. `// check if this is it` looks like a leftover TODO. Remove it (or replace with the resolved decision).
- **`src/webapp/.../monaco/registerInlineConstantCompletions.ts:107` — silent fetch failure**. `options.fetchConstants(prefix).catch(() => [] as Constant[])` swallows API errors. The user types in `texts.X.code`, the API fails, and they see "no constants" with no signal. Surface the failure: at minimum `console.error`, ideally a non-selectable completion item with `detail: i18n.t("Failed to load constants")`.
- **`src/webapp/.../monaco/registerInlineConstantCompletions.ts:25-35` — typing**. The function signature uses `any` for `monacoEditor`, the model and the `position` are `any`, and the return objects are `any`. Use the types exported by `monaco-editor` (`editor.IStandaloneCodeEditor`, `languages.CompletionItem`, etc.). The dead-code fallback `commandId: registeredCommandId ?? commandId` (the locally-generated `commandId` is never registered with `addCommand`, so the fallback would never fire a real command) should also be removed once typing forces `addCommand` to return a `string`.
- **`src/webapp/.../Editor.tsx:13,21` — `useState<any>()`**. `setEditorInstance` and the `range` parameter on `EditorApi.insertAtRange` use `any`. Reuse `editor.IStandaloneCodeEditor` and `editor.IRange` instead of inventing a parallel `MonacoRange` shape (also duplicated in `Configurator.tsx` as `PendingRange`). Extract a small `monaco/types.ts` if you want to keep the import surface narrow.
- **`src/webapp/.../CreateConstantDialog.tsx:189` — duplicate-code detection by regex on the error message**. `/code/i.test(message) && /exist|duplicate|unique/i.test(message)` will silently misclassify on locale changes or DHIS2 message tweaks. The right place is the `errorReports[].errorCode` returned in `typeReports`, which `Stats.errorMessage` flattens away. Two options: (a) enrich `Stats` with structured field/error data, or (b) have `CreateConstantUseCase` throw a typed error that includes the field. Acceptable for v1, but mark as a known fragility (TODO + follow-up issue).
- **`src/data/common/Dhis2DataStoreDataForm.ts:626` — default value vs codec contract**. `defaultDataStoreConfig.prefix = ""` while the codec types it `optional(string)`. Either align the default to `undefined` (and let consumers `?? ""`), or change the codec to a non-optional `string` with a default. Right now the docs say "absent ⇒ no filter" but `defaultDataStoreConfig` materialises an empty string.

## 2. Recommendations non blocking

- **`extractRootPrefix.ts` has no unit tests**. Add cases for: prefix absent, prefix as a non-string, escaped characters in the value, prefix in the middle of mid-typed (invalid) JSON. The regex fallback's `.replace(/\\(.)/g, "$1")` un-escapes too aggressively (`\n` → `n`); fine for ASCII prefixes but worth a comment narrowing the supported character set.
- **`jsonPathAtCursor.ts` — extra walker tests**. The 7 existing cases cover the happy path. Hand-rolled JSON traversal benefits from tests for: cursor inside a string with escaped quotes (`"a\"b"`), `code` inside an array element under `texts`, very deeply nested objects, and trailing comma / mid-typed malformed JSON.

xurxodev (approved):
thanks @deeonwuli 

2. Recommended non-blocking

The IA agents configuration is duplicated, but I think this is a topic to treat in the coding dojo meeting

src/domain/common/entities/User.ts
```diff
@@ -8,4 +8,11 @@ export interface User {
     orgUnits: OrgUnit[];
     userRoles: NamedRef[];
     userGroups: NamedRef[];
+    authorities: ReadonlyArray<string>;
+}
+
+export const CONSTANT_ADD_AUTHORITY = "F_CONSTANT_ADD";
+
+export function canCreateConstants(user: User): boolean {
+    return user.authorities.includes("ALL") || user.authorities.includes(CONSTANT_ADD_AUTHORITY);
```
eperedo: Authorities are a DHIS concept. I think we should include this validation in the data layer and just include`bool: canCreateConstant` property in the User entity.

eperedo (changes requested):
Looks good, thanks @deeonwuli!

A comment about clean arch and can you check the version in the `package.json` file. In [development we have 1.5.3](https://github.com/EyeSeeTea/d2-autogen-forms/blob/development/package.json#L4) but in this branch is 1.5.1.

eperedo (approved):

PR #183, opened 2026-04-27
[refactor]: Convert side-effects from useMemo to useEffect in iFramePortal

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869d1mfbp

### :memo: Implementation
- [x] Replaced `useMemo` with `useEffect` + `useState` in `IframeEscapePortal`.                    
- [x] Extract `getTopAccessibleWindow()` into `src/webapp/utils/topAccessibleWindow.ts` so the parent-window resolution logic is shared instead of duplicated.           
- [x] Added a clarifying comment in `useApplyRules.getValueAndVerifyCondition` explaining why visibility/disabled/enabled rules evaluate against the form's base period rather than the section's offset period.           
         
### :art: Screenshots
Screenshot 2026-04-27 at 16 51 23

- Check if the validation rules popup is at the bottom right of the current viewable screen and not the entire page in the new data entry app.

#869d1mfbp

### DIFF
--- a/src/webapp/components/iframe-escape-portal/IframeEscapePortal.tsx
+++ b/src/webapp/components/iframe-escape-portal/IframeEscapePortal.tsx
@@ -1,8 +1,10 @@
-import React, { useMemo } from "react";
+import React, { useEffect, useRef, useState } from "react";
 import ReactDOM from "react-dom";
 import { StyleSheetManager } from "styled-components";
 import { StylesProvider, jssPreset } from "@material-ui/core/styles";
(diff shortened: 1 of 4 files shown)


MatiasArriola (approved):
Good refactor, thanks @deeonwuli!

MiquelAdell (approved):
comes from the analysis here:
https://github.com/EyeSeeTea/d2-autogen-forms/pull/174

PR #185, opened 2026-04-30
 [fix]: Apply rules using trigger section/row period

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869cmxcc8

### :memo: Implementation
[PR #177](https://github.com/EyeSeeTea/d2-autogen-forms/pull/177) made rule condition lookups always use the form's base period. That broke rules whose trigger DE lives in a section with a period offset: in PAHO country TUB sections 6.2 / 6.4 (`startOffset/endOffset: -1`) the trigger's value sits at the cohort year, not the form year, so the lookup found nothing and the dependent rows in 6.3 / 6.5 stayed disabled.
                                                                                                                                                                                                                           
`getApplicablePeriod` now resolves the period from the row period when the trigger's section has data there (so multi-period sections evaluate per-row), otherwise the form's base period.

### :art: Screenshots

https://github.com/user-attachments/assets/6aab041f-df7b-402f-9dc6-1761c3f8396a


#869cmxcc8

### DIFF
--- a/src/webapp/reports/autogenerated-forms/hooks/useApplyRules.ts
+++ b/src/webapp/reports/autogenerated-forms/hooks/useApplyRules.ts
@@ -23,7 +23,7 @@ type UseApplyRulesProps = {
 
 type FormRuleOptions = Omit<UseApplyRulesProps, "dataElementTotalRules" | "dataElement"> & {
     rules: Rule[];
-    dataElementTotalRule: Maybe<DataElementTotalRule | SectionTotalRule>;
+    dataElementTotalRule: Maybe<DataElementTotalRule>;
 };
 
 type UseApplyRulesReturn = { isVisible: boolean; isDisabled: boolean };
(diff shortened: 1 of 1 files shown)


MatiasArriola (changes requested):
Good refactor for `getValueAndVerifyCondition` to receive an options object instead a lot of positional parameters. Also, `getApplicablePeriod` is a nice and well-focused abstraction. The intent aligns well with the PR description. 

Only some comments that might be worth reviewing: 
- `getValueAndVerifyCondtion`: We are using `period` instead of  `applicablePeriod` in the `evaluateTotalRule` call. Is that correct? Maybe this is just a smell, but wanted to raise the question. I think we should confirm and test around this, and if no change is needed, add comment on why `period` is used.
- `FormRuleOptions` type includes `dataElementTotalRule: Maybe<DataElementTotalRule | SectionTotalRule>;` but `SectionTotalRule seems to be not needed

Thanks @deeonwuli!

deeonwuli:
Good catch @MatiasArriola. There aren't currently any total rules that span elements from sections with different periods, so the existing behavior hasn't caused issues in practice. That said, it is a latent bug, `evaluateTotalRule` should be calling `getApplicablePeriod` per related data element rather than applying a single period uniformly. We'll make that fix.

MatiasArriola (approved):

PR #186, opened 2026-05-05
fix(toggleMultiple): avoid vacuous-truth disable when no orgUnit toggles

MiquelAdell:
## Task:
https://app.clickup.com/t/869d5pkq2

## Summary
- Fix regression where every cell in a `toggleMultiple` section was disabled when `logicalOperator: "AND"` was combined with only `dataElement`-type toggles (no orgUnit toggles).
- `isToggleMultipleDeDisabled` filtered `toggleDataElements` down to orgUnit entries matching the data element, then evaluated with `.every()`. With an empty array, `[].every(...)` is vacuously `true`, locking the section regardless of toggle state.
- Short-circuits to `false` when there are no matching orgUnit toggles, so the function only ever returns `true` because an actual orgUnit-disable rule applied.

Regression introduced in #148.

## Repro
Reported by Ryan (WHO MAL): the WMR `MAL_WMR_LAB_DBL_COUNT_GRID` section (`Double counted figures` under `TOTAL MALARIA CASES`) had all `Public/Private/Community/Combined` cells for `MAL_MICR_RDT_DOUBLE_COUNT_TEST` and `MAL_MICR_RDT_DOUBLE_COUNT_POS` locked even when the upstream Yes/No questions (`MAL_MICR_RDT_DOUBLE_COMB`, `MAL_MICRDTDOUBLECOUNTFIGURESKNOWNCOMB`) were Yes. The deployed datastore configures this section with `logicalOperator: "AND"` and two `dataElement`-type conditions, hitting the vacuous-truth path.

## Test plan
- [x] Existing unit tests in `src/domain/common/entities/__tests__/ToggleMultiple.spec.ts` pass.
- [x] Verified live (HMR) on the WMR form (`dataSetId=CWuqJ3dtQC4`, `orgUnit=hmZE3mVAZFf`, `period=2025`): the 8 number inputs in `Double counted figures` are now `disabled: false` and editable.
- [x] Verified TUB Annual section 6 (`dataSetId=XGn6mrO2HU8`) is unaffected — its sections use `logicalOperator: "OR"` with orgUnit-type toggles, neither condition triggers the new early-return.


MatiasArriola (approved):
The fix is correct and safe.

`[].every(() => false)` is `true`, so we were incorrectly returning `true` there.

Thanks @MiquelAdell!

PR #187, opened 2026-05-13
feat(configurator): add Install Custom Form button

MiquelAdell:
## Summary

- Adds an **Install Custom Form** button to the Configurator that pushes the bundled `custom-data-form.html` directly into a dataset's DHIS2 data-entry form via `POST /api/dataSets/{id}/form`
- Detects existing install state on dataset selection (not installed / installed vX.Y.Z / outdated / unknown version) and shows a labelled chip: **Custom form:** `installed (v1.5.3)`
- Install section positioned below the editor, grouped with Save/Clear/Delete actions
- Version marker (`<!-- d2-autogen-forms-custom-form vX.Y.Z -->`) is injected at build time so installs are traceable
- `build-configurator` and new `start-configurator` scripts bundle the form HTML into the zip/public folder automatically

Closes ClickUp task [#869d7dcpm](https://app.clickup.com/t/869d7dcpm)

## Test plan

- [x] TypeScript compiles cleanly (`yarn tsc --noEmit`)
- [x] All 72 existing tests pass (`yarn test`)
- [x] Manual smoke test on `who-dev-41` (DHIS2 v2.41): selected MAL - WMR Form, install state detected as "unknown version", re-installed, chip updated to "installed (v1.5.3)"

## Screenshots
image


src/webapp/reports/autogenerated-forms-configurator/hooks/useInstallCustomForm.ts
```diff
@@ -0,0 +1,73 @@
...
+
+        setIsInstalling(true);
+        try {
+            await compositionRoot.customForm.install(dataSetId, bundledHtml);
+            snackbar.success(i18n.t("Custom form installed successfully"));
+            const state = await compositionRoot.customForm.getInstallState(dataSetId, bundledHtml);
+            setInstallState(state);
+        } catch (error: unknown) {
+            const message = error instanceof Error ? error.message : String(error);
+            snackbar.error(i18n.t(`Error installing custom form: ${message}`));
```
eperedo: it should be:

```ts
snackbar.error(i18n.t("Error installing custom form: {{message}}", { message, nsSeparator: false }));
```

docs/proposals/install-custom-form-button.md
```diff
@@ -0,0 +1,305 @@
...
+
+- **First-time install** — writes `custom-data-form.html` into the dataset's
+  custom data-entry form.
+- **Re-install / update** — if the dataset already has a custom form, the
+  configurator shows the bundled-form version vs. the installed-form version
+  (extracted from a version marker embedded in the HTML — see
+  [Technical approach §3](#3-detect-installed-form-state)) and lets the admin
+  overwrite after a confirmation step.
+
+State display: an indicator on the dataset row in the dataset list — *Custom
```
eperedo, line 49: Not sure if I'm missing something but according to this part of the proposal I was expecting to have a status per dataSet in the selector.

Image

docs/proposals/install-custom-form-button.md
```diff
@@ -0,0 +1,305 @@
...
+
+DHIS2 stores a dataset's custom form on `dataSet.dataEntryForm` (entity
+`DataEntryForm` with field `htmlCode`).
+
+**Authority required:** write access to the target `DataSet`. There are no
+separate `F_DATAENTRYFORM_*` authorities — dataset write access is sufficient
+for both creating and updating `dataEntryForm`.
+
+**Preferred approach — single metadata request** (handles create and update
+in one call, both for first-time install and re-install):
```
eperedo, line 130: The implementation is using a different endpoint: `/api/dataSets/{dataSetId}/form`. Not wrong, but just to point out that the proposal was ignored in the implementation.

src/domain/autogenerated-forms-configurator/usecases/InstallCustomFormUseCase.ts
```diff
@@ -0,0 +1,10 @@
+import { Id } from "../../common/entities/Base";
+import { CustomFormRepository } from "../repositories/CustomFormRepository";
+
+export class InstallCustomFormUseCase {
+    constructor(private customFormRepository: CustomFormRepository) {}
+
+    async execute(dataSetId: Id, htmlCode: string): Promise<void> {
+        await this.customFormRepository.install(dataSetId, htmlCode);
```
eperedo: We don't have the permission validation here.

From the proposal:

```shell
**Authority required:** write access to the target `DataSet`. There are no
separate `F_DATAENTRYFORM_*` authorities — dataset write access is sufficient
for both creating and updating `dataEntryForm`.
```

We can get this information from the access field for each dataSet:

Image

and add a new field in the dataset Entity

```ts
export type DataSet = CodedRef & {
    id: Id;
    configExists: boolean;
    hasCustomForm: boolean;
   // new property
    canBeModified: boolean;
};
```

Then add the validation in the usecase

```ts
async execute(dataSetId: Id, htmlCode: string): Promise<void> {
   // fetch dataSet
   const dataSet = await this.dataSetRepository.get(dataSetId);
   if (dataSet.canBeModified) {
        await this.customFormRepository.install(dataSetId, htmlCode);
   } else {
        throw Error("Forbidden");
   }

}
```

src/domain/autogenerated-forms-configurator/repositories/CustomFormRepository.ts
```diff
@@ -0,0 +1,6 @@
+import { Id } from "../../common/entities/Base";
+
+export interface CustomFormRepository {
+    getInstalledHtml(dataSetId: Id): Promise<string | null>;
```
eperedo: Normally we return an entity from the repository instead a primitive type.

A `CustomFormRepository` should return a `CustomForm` entity:

```ts
type CustomForm = {
  template: string;
  status: CustomFormInstallState;
}

export interface CustomFormRepository {
    get(dataSetId: Id): Promise<CustomForm>;
}
```

src/webapp/reports/autogenerated-forms-configurator/Configurator.tsx
```diff
@@ -62,6 +88,33 @@ const AutogeneratedFormConfigurator: React.FC = () => {
...
 
+            {dataSetCode && (
+                <InstallSection>
+                    <CustomFormLabel>{i18n.t("Custom form:")}</CustomFormLabel>
+                    <InstallStateChip installState={installState} />
+                    <Tooltip
+                        title={
+                            !bundledVersion
+                                ? i18n.t(
+                                      "Bundled custom form not found. Run yarn start-configurator to make it available."
```
eperedo: Not wrong, but to me this looks like a message that could be understood only by developers or technical users only.

src/data/autogenerated-forms-configurator/Dhis2DataSetRepository.ts
```diff
@@ -30,15 +30,31 @@ export class Dhis2DataSetRepository implements DataSetRepository {
             })
             .getData();
 
-        return dataSets.map(dataSet => ({ name: dataSet.name.trim(), code: dataSet.code.trim() }));
+        return dataSets.map(dataSet => ({
```
eperedo: I don't understand why do we need the new type `DataSetMetadata` instead of the existing `D2DataSet` from MetadataPick.

Here we can directly `return dataSets;` and then calculate the `hasCustomForm` value in the other method `get`

```ts
return d2DataSets.map(dataSet => {
    const configExists = autogenDatastoreKeys.includes(dataSet.code);

    return {
        id: dataSet.id,
        name: dataSet.name.trim(),
        code: dataSet.code.trim(),
        hasCustomForm: dataSet.formType === "CUSTOM",
        configExists,
    };
});
```

docs/proposals/install-custom-form-button.md
```diff
@@ -0,0 +1,305 @@
...
+
+Where `vX.Y.Z` matches `package.json` `version`. To determine state, the
+configurator reads `dataSet.dataEntryForm.htmlCode`, parses the marker, and
+compares to the bundled version. Three states:
+
+| State | Condition |
+|---|---|
+| Not installed | `dataSet.dataEntryForm` is null |
+| Installed (current) | marker version === bundled version |
+| Outdated | marker version !== bundled version (or marker absent) |
```
eperedo, line 178: In the implementation there's a four state `unknown-version` when the marker is not present.

docs/proposals/install-custom-form-button.md
```diff
@@ -0,0 +1,305 @@
...
+The App Hub submission was the *configurator* zip, so the in-DHIS2 display
+name is already correct. What's left for the rebrand:
+
+- **README** — lead with "This is the configurator app for the
+  autogenerated-forms custom form. Install it from the App Hub to manage
+  which datasets get the form, and click *Install custom form* to deploy the
+  renderer." Today's README emphasises the legacy reports app and only
+  mentions the configurator at the end.
+- **App Hub submission description** — aligned to the same framing.
+- **`appVersions.configurator`** — currently `1.5.0`, behind
```
eperedo, line 76: appVersions.configurator is still 1.5.0 in the package.json file

eperedo (changes requested):
Thanks @MiquelAdell, tested on dev and it works well. Some observations about clean arch. and some implementations that are not aligned to the proposal.

src/webapp/reports/autogenerated-forms-configurator/hooks/useInstallCustomForm.ts
MiquelAdell: Claude: Fixed. Now using `i18n.t("Error installing custom form: {{message}}", { message, nsSeparator: false })`.

docs/proposals/install-custom-form-button.md
MiquelAdell: Claude:  Fetching `htmlCode` for every dataset upfront would be expensive. The full state (installed/outdated/version) shows only after selecting a dataset. We do already have `hasCustomForm` per dataset from the initial call, which could be a lightweight dropdown indicator if that seems useful. Discussed with PM — this diverges from the proposal and it's documented in the PR.

Miquel: I confirm that ^^

docs/proposals/install-custom-form-button.md
MiquelAdell: Claude: Switched to `this.api.metadata.post({ dataEntryForms: [...], dataSets: [...] })`. That's the pattern `ConstantD2Repository` uses, so now it's consistent. The `/dataSets/{id}/form` endpoint worked but was out of step with the rest of the codebase.

Miquel: I'm ok with deviating from the plan, the code wins in this case

src/domain/autogenerated-forms-configurator/usecases/InstallCustomFormUseCase.ts
MiquelAdell: Claude: Added. `DataSet` now has `canBeModified: boolean` from `access.data.write`. `DataSetRepository` gets a new `getById(id)`. The use case calls it before installing and throws if `canBeModified` is false. On the UI side, the button is disabled with a tooltip ("You do not have write access to this dataset.") when the dataset isn't writable.

src/domain/autogenerated-forms-configurator/repositories/CustomFormRepository.ts
MiquelAdell: Claude: Created `CustomForm`: `{ id: string | null; htmlCode: string | null }`. Went with raw data over `{ template, status }` — the use case is a better place for computed state, and `id` is needed to reuse the existing `DataEntryForm` on re-installs rather than creating a second one. Decision is in the PR description.

Miquel: @eperedo, if I understand this correctly Claude pushed against your decision and I told it to go ahead and propose its solution as the reasoning, on the surface, made sense. Please feel free to push back if needed and I'll accept your final say in that as this particular point I do not know enough to decide for one or the other

src/webapp/reports/autogenerated-forms-configurator/Configurator.tsx
MiquelAdell: Claude: Fixed. Now says "Custom form bundle not available. Please reinstall the application." Added a second tooltip for the access case: "You do not have write access to this dataset."

src/data/autogenerated-forms-configurator/Dhis2DataSetRepository.ts
MiquelAdell: Claude: Removed. `getMetadata()` now returns `D2DataSet[]` straight from `MetadataPick`. The mapping (`name.trim()`, `hasCustomForm`, `canBeModified`, `configExists`) happens in `get()` and the new `getById()`.

docs/proposals/install-custom-form-button.md
MiquelAdell: Claude: Merged into `outdated`. Three states now: `not-installed | installed | outdated`. No marker returns `{ type: "outdated", installedVersion: null, bundledVersion }`. Chip shows "outdated (unknown → v1.5.3)".

docs/proposals/install-custom-form-button.md
MiquelAdell: Claude: Bumped to 1.5.3.

MiquelAdell:
Claude: All addressed in this commit. Two decisions documented in the PR: `CustomForm` uses `{ id, htmlCode }` rather than `{ template, status }` because `id` is needed by the metadata API on re-installs and computed state fits better in the use case. For the selector, fetching `htmlCode` per dataset upfront is too expensive, so the full state stays in the detail view — talked to PM, it's fine.

Miquel: I reviewed interactively the proposals for fixes from Claude and then approved drafts for each of the responses. Also marked here what part of the response came from Claude and what from me.
@eperedo  From my viewpoint everything looks "reasonable" but please let me know if we need more polishing or you disagree with any of the decisions.

src/domain/autogenerated-forms-configurator/repositories/CustomFormRepository.ts
eperedo: Yes, my bad. the CustomForm entity in my comment was just an example not the exact implementation. I agree keeping status a calculated state is ok in the usecase.

eperedo (changes requested):
@MiquelAdell I see the comments from Claude, but no code changes. Did it forget to push the changes?

src/data/autogenerated-forms-configurator/Dhis2DataSetRepository.ts
```diff
@@ -10,17 +11,45 @@ export class Dhis2DataSetRepository implements DataSetRepository {
...
+            configExists: autogenDatastoreKeys.includes(dataSet.code.trim()),
+        };
     }
 
-    private async getMetadata(): Promise<DataSetMetadata[]> {
+    private async getMetadata(): Promise<D2DataSet[]> {
         const { objects: dataSets } = await this.api.models.dataSets
             .get({
-                fields: dataSetFields,
+                fields: { ...dataSetFields, access: true } as unknown as typeof dataSetFields,
```
eperedo: we can add the access: true in the variable `dataSetFields` then we don't need to use the cast `as unknown as typeof dataSetFields`

src/data/autogenerated-forms-configurator/Dhis2DataSetRepository.ts
```diff
@@ -46,15 +70,8 @@ const dataSetFields = {
...
     dataSets: { fields: typeof dataSetFields };
 }>["dataSets"][number];
 
-type DataSetMetadata = {
-    id: string;
-    name: string;
-    code: string;
-    hasCustomForm: boolean;
-};
+type D2DataSetWithAccess = D2DataSet & { access?: { data?: { write?: boolean } } };
```
eperedo: We don't need a new type `D2DataSetWithAccess` if we include the access field in the `dataSetFields` variable

src/data/autogenerated-forms-configurator/Dhis2DataSetRepository.ts
```diff
@@ -10,17 +11,45 @@ export class Dhis2DataSetRepository implements DataSetRepository {
...
         const autogenDatastoreKeys = await this.api.dataStore(Namespaces.D2_AUTOGEN_FORMS).getKeys().getData();
 
-        return d2DataSets.map(dataSet => {
-            const configExists = autogenDatastoreKeys.includes(dataSet.code);
+        return d2DataSets.map(dataSet => ({
+            id: dataSet.id,
+            name: dataSet.name.trim(),
+            code: dataSet.code.trim(),
+            hasCustomForm: dataSet.formType === "CUSTOM",
+            canBeModified: (dataSet as D2DataSetWithAccess).access?.data?.write ?? false,
```
eperedo: We need to check for the value in `dataSet.access.write` which is the one related to metadata access

src/data/autogenerated-forms-configurator/Dhis2DataSetRepository.ts
```diff
@@ -10,17 +11,45 @@ export class Dhis2DataSetRepository implements DataSetRepository {
...
+            id: string;
+            name: string;
+            code: string;
+            formType: string;
+            access: { data: { write: boolean } };
+        };
+
+        const dataSet = await this.api
+            .get<DataSetResponse>(`/dataSets/${id}`, { fields: "id,name,code,formType,access" })
+            .getData();
```
eperedo, line 30: We normally use `this.api.models.dataSets.get({ fields: dataSetFields, filter: { id: { eq: id }} })` so we can reuse the d2-api schemas.

src/data/autogenerated-forms-configurator/Dhis2DataSetRepository.ts
```diff
@@ -10,17 +11,45 @@ export class Dhis2DataSetRepository implements DataSetRepository {
...
+        const autogenDatastoreKeys = await this.api.dataStore(Namespaces.D2_AUTOGEN_FORMS).getKeys().getData();
 
-            return { ...dataSet, configExists };
-        });
+        return {
+            id: dataSet.id,
+            name: dataSet.name.trim(),
+            code: dataSet.code.trim(),
+            hasCustomForm: dataSet.formType === "CUSTOM",
+            canBeModified: dataSet.access?.data?.write ?? false,
```
eperedo: We should check for `dataSet.access.write` for metadata access

eperedo (changes requested):
We're close just some typescript and d2-api minor improvements

src/data/autogenerated-forms-configurator/Dhis2DataSetRepository.ts
MiquelAdell: Claude: Added access: true to dataSetFields and dropped the as unknown as typeof dataSetFields cast.

src/data/autogenerated-forms-configurator/Dhis2DataSetRepository.ts
MiquelAdell: Claude: Refactored getById() to use d2-api models pattern with dataSetFields constant. Removed the unused DataSetResponse type. Typechecks without errors.

src/data/autogenerated-forms-configurator/Dhis2DataSetRepository.ts
MiquelAdell: Claude: Fixed. The distinction here is that `access.write` checks **metadata write** permission (can the user modify the dataset definition?), while `access.data.write` checks **data entry** permission (can the user enter data?). For determining whether a dataset form can be modified, we need the metadata permission. Also, since `access: true` is now in `dataSetFields`, the d2-api schema types `access` directly on `D2DataSet`, so the cast is unnecessary.

src/data/autogenerated-forms-configurator/Dhis2DataSetRepository.ts
MiquelAdell: Claude: Fixed. Changed getById() to use dataSet.access.write matching the get() method. Typecheck passes.

src/data/autogenerated-forms-configurator/Dhis2DataSetRepository.ts
MiquelAdell: Claude: Unused type alias removed. Verified no other references exist with grep, typecheck passes, linting clean.

MiquelAdell:
@eperedo Thank you for your review. All changes addressed

eperedo (approved):

PR #188, opened 2026-05-14
fix(delete-rule): scope deletions to the trigger value's period

MiquelAdell:
## Summary

- The delete-rule sweep was matching the form's value store only on data element code and org unit, so values from referenced periods (loaded into the store for historical columns) were swept into the same DELETE batch as the current period. Toggling a trigger on period N could silently wipe a value the user had previously saved on period N-1.
- Fix scopes deletions to the trigger value's own period. Using the trigger's period (rather than the form's base period) correctly handles offset sections where trigger and targets share an offset.
- COC is intentionally left unconstrained: triggers are typically single-COC and target the full disaggregation of a DE (e.g. `TUB_agegroup_option` clearing all COCs of `TUB_newrel`).

Refs ClickUp [869da0yxb](https://app.clickup.com/t/869da0yxb).

## Test plan

Verified locally against `https://extranet.who.int/dhis2-dev` with the GTB dataset (`XGn6mrO2HU8`, OU `oofyLUJJ6Vy`):

- [x] Pre-existing 2024 value for `TUB_newinc_con` (`N9JwlMJCg2P` / `Xr12mI7VPn3`) survives a change to `TUB_screen_data_available` on period 2025. DELETE payload contains only `period:"2025"`.
- [x] Changing `TUB_agegroup_option` (single-COC trigger) still deletes `TUB_newrel` values across non-default COCs.
- [x] `yarn tsc --noEmit` passes.

### DIFF
--- a/src/webapp/reports/autogenerated-forms/hooks/useDeleteDataValues.ts
+++ b/src/webapp/reports/autogenerated-forms/hooks/useDeleteDataValues.ts
@@ -45,7 +45,9 @@ export function useDeleteDataValues(props: UseDeleteRulesProps): UseDeleteRuleRe
                     _(dataFormInfo.data.values.store)
                         .filter(
                             dataValue =>
-                                dataValue.dataElement.code === deId && dataValue.orgUnitId === dataFormInfo.orgUnitId
+                                dataValue.dataElement.code === deId &&
+                                dataValue.orgUnitId === dataFormInfo.orgUnitId &&
+                                dataValue.period === dataValueCb.period
                         )
                         .compact()
                         .value()


MiquelAdell:
Context on the history of this filter, for reviewer benefit:

- Added in 19250b5 (2026-01-21, @gqcorneby) as `dataValue.period === (period || dataFormInfo.period)`.
- Removed in edfad89 (2026-01-26, @gqcorneby) with no commit-message context.

@gqcorneby — I think we dropped this because we were targeting changes in 2024 from the 2025 form for sections that were offset, and this was failing. I guess the error was to just remove it instead of making it dependent on the period of the data element.

gqcorneby:
> @gqcorneby — I think we dropped this because we were targeting changes in 2024 from the 2025 form for sections that were offset, and this was failing. I guess the error was to just remove it instead of making it dependent on the period of the data element.

Hi @MiquelAdell! I tried to recall as best I could and I think you are right but inverted: in an offset section the trigger fires at the section's offset period (e.g. 2024), but the old filter compared against the form's base period (2025), so the sweep matched nothing and the rule silently no-op'd on offsets. Dropping the filter unblocked that case, but it overcorrected and deleted historical reference. Apologies for missing this, I think I may have missed the fact that historical reference values were sitting in the same `values.store`.

Approving as the fix looks correct.

gqcorneby (approved):

PR #189, opened 2026-05-14
[feature]: Replace custom header in configurator with DHIS2 HeaderBar

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869d9dka9

### :memo: Implementation
Replaces the custom configurator header with the standard DHIS2 `HeaderBar` component from `@dhis2/ui`.

- [x] Added a `HeaderBar` wrapper in `App.tsx` that renders the DHIS2 header bar when running as a standalone app
- [x] The header is suppressed when the app runs inside an iframe (e.g. embedded in DHIS2 data entry) using `window.self !== window.top`, since the parent shell already provides one
- [x] Added missing module declarations for `@dhis2/ui` and `*.css` imports to `declarations.d.ts`

### :art: Screenshots
Screenshot 2026-05-14 at 12 41 58

- Open the configurator directly — the DHIS2 header bar should appear at the top
- Open the autogenerated form embedded in DHIS2 data entry — no duplicate header should appear

#869d9dka9


MiquelAdell (approved):
looks good, thanks!

eperedo (approved):

PR #192, opened 2026-05-20
[fix]: Replace invalid-XML control chars from data values on save

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869dbc90g

### :memo: Implementation
The client report surfaced a Bulk Load Excel export that failed to open in MS Excel (`error237400_01.xml`: "Illegal xml character. Line 395, column 174" in `/xl/sharedStrings.xml`). 
Root cause: an invisible **vertical tab (`\x0B`)** had been entered into a TEXT data value. XLSX is XML 1.0; XML 1.0 forbids most C0 control characters, so Excel rejected the file. LibreOffice silently strips these on save, which is why a LibreOffice round-trip "fixed" the file.

This PR sanitizes data values at the save boundary in autogen-forms so the same characters never get persisted again:

**Out of scope (sibling task in Bulk Load):** historical bad values already in DHIS2 and values entered through other paths (DHIS2 web data entry, imports) still need a defensive strip on the Bulk Load xlsx export side. [Bulk Load PR 399](https://github.com/EyeSeeTea/Bulk-Load/pull/399)

  #869dbc90g

### DIFF
--- a/src/data/common/Dhis2DataValueRepository.ts
+++ b/src/data/common/Dhis2DataValueRepository.ts
@@ -14,7 +14,7 @@ import {
 import { DataValueRepository, DataElementRefType } from "../../domain/common/repositories/DataValueRepository";
 import { D2Api, DataValueSetsDataValue } from "../../types/d2-api";
 import { promiseMap } from "../../utils/promises";
-import { stripInvalidXmlChars } from "../../utils/string";
+import { replaceInvalidXmlChars } from "../../utils/string";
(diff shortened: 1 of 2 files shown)


eperedo (approved):
Looks good to me, thanks @deeonwuli

MiquelAdell (approved):
tested by inputing the result of this command `printf 'TEST1\vTEST2' | pbcopy` it works

PR #197, opened 2026-05-25
fix(data): use mergeMode=MERGE when installing custom form to prevent data loss

MiquelAdell:
### :pushpin: References

- **Issue:** Regression introduced by #195 / PR #195

### :memo: Implementation

The metadata POST in `Dhis2CustomFormRepository.install()` was defaulting to `mergeMode=REPLACE`, which replaces the **entire** dataset object in DHIS2 with the partial payload we send (only `id`, `name`, `shortName`, `periodType`, `dataEntryForm`). This erased all other dataset properties: `translations`, `sharing`, `dataSetElements`, `sections`, `organisationUnits`, `legendSets`, etc.

Fix: pass `{ mergeMode: "MERGE" }` as the second argument to `api.metadata.post()`. With `MERGE` mode, DHIS2 only updates the fields present in the payload and leaves everything else intact.

The `mergeMode: "MERGE"` option is already used in `src/scripts/post-metadata.ts` and is typed as a valid `PostOptions` field in `@eyeseetea/d2-api`.

### :art: Screenshots

`before.json` (dataset before install) vs `after.json` (dataset after install with the bug) confirmed data loss in: `translations`, `sharing.userGroups`, `sharing.public`, `legendSets`, `dataSetElements`, `sections`, `organisationUnits`.

After this fix, only the `dataEntryForm` reference is updated on the dataset; all other properties are preserved.

1. Note the current state of a dataset (sharing, dataSetElements, sections, organisationUnits, translations).
2. Install a custom form via the configurator.
3. Verify the dataset still has its original sharing, dataSetElements, sections, organisationUnits, and translations.


src/scripts/verify-form-install.ts
```diff
@@ -0,0 +1,117 @@
...
+                : aStr?.slice(0, 80);
+            diffs.push(`CHANGED: ${fullPath}\n  before: ${bSummary}\n  after:  ${aSummary}`);
+        }
+    }
+
+    return diffs;
+}
+
+async function fetchDataSet(api: D2Api, dataSetId: string): Promise<Record<string, unknown>> {
+    const response = await api
```
eperedo: we can use the `models.api.dataSets({ fields: $all, filter: ... })` or `models.api.metadata({ dataSets: { fields: $all, filter: ... } })` to get the D2DataSet schema from d2-api and avoid the unknown type.

src/scripts/verify-form-install.ts
```diff
@@ -0,0 +1,117 @@
...
+const SAMPLE_HTML = "<div>d2-autogenforms verification form</div>";
+
+function getEnv(name: string): string {
+    const val = process.env[name];
+    if (!val) throw new Error(`Missing required env var: ${name}`);
+    return val;
+}
+
+function flatDiff(before: Record<string, unknown>, after: Record<string, unknown>, path = ""): string[] {
+    const diffs: string[] = [];
```
eperedo: I know this is just a script so not a big deal, but we try to follow functional programming and avoid mutating variables.

eperedo (changes requested):

MiquelAdell:
Thanks for the feedback @eperedo
 
Both points addressed in 54ebc69:

  - Mutable accumulator: refactored flatDiff to use flatMap, with each branch returning a string array instead of pushing into a shared accumulator.
  - unknown type: replaced api.get<Record<string, unknown>> with api.models.dataSets.get({ fields: { $all: true }, filter: { id: { eq: dataSetId } } }). The return type is now
  SelectedPick<D2DataSetSchema, { $all: true }> — fully typed from d2-api. The cast to Record<string, unknown> at the flatDiff call site is intentional since that function does a
  recursive generic diff.

eperedo (approved):

PR #198, opened 2026-06-02
feat(configurator): prompt to create default section when dataset has none

MiquelAdell:
### :pushpin: References

- **ClickUp:** [make autogen config create default section in the metadata if needed](https://app.clickup.com/t/869dhgfra)

### :memo: Implementation

When saving a DataStore config for a dataset that has no sections, the data entry form shows a dead-end "There are no sections in this data set." message. This PR adds a guard at save time.

After a successful config save, the configurator checks whether the target dataset has any sections. If it has none, a confirmation dialog appears:

> "The target dataset does not have any section. Sections are needed for d2-autogen-forms to work. Do you want to create a default section with all data elements?"

- **Yes** — creates a DHIS2 section named "Default" containing all the dataset's data elements via `POST /api/metadata`, then shows a success snackbar.
- **No** — closes the dialog, no changes made.

Datasets that already have sections are unaffected (no dialog shown).

**Changes by layer:**
- `domain/entities/DataSet.ts` — added `DataElementRef` and `DataSetDetail` types
- `domain/repositories/DataSetRepository.ts` — added `getWithElements` and `createDefaultSection` to interface
- `domain/usecases/` — two new use cases: `GetDataSetWithElementsUseCase`, `CreateDefaultSectionUseCase`
- `data/Dhis2DataSetRepository.ts` — implemented both methods; `generateUid()` for the section UID, `api.metadata.post` to persist
- `compositionRoot.ts` — wired new use cases under `dataSet` namespace
- `hooks/useConfigurator.ts` — post-save section check, dialog state, confirm/cancel handlers
- `Configurator.tsx` — renders the `ConfirmationDialog`
- `i18n/` — updated `.pot` and `.po` files with 3 new strings

### :art: Screenshots

_Dialog shown after saving config for a dataset with no sections:_

> "Dataset has no sections — The target dataset does not have any section. Sections are needed for d2-autogen-forms to work. Do you want to create a default section with all data elements? [No] [Yes]"

1. Use a DHIS2 instance where a dataset exists with **no sections** (create one in Maintenance if needed).
2. Open the Autogenerated Forms Configurator and select that dataset.
3. Load or write a valid config JSON and click **Save**.
4. Confirm the dialog appears after the "Configuration saved" snackbar.
5. Click **No** → dialog closes, no section created (verify in Maintenance).
6. Repeat and click **Yes** → "Default section created" snackbar appears; verify the section exists in Maintenance with all data elements.
7. Repeat with a dataset that **already has sections** → no dialog should appear.

Test instance: https://play.im.dhis2.org/stable-2-42-5/ (user: `system`, password: `System123`)


src/domain/autogenerated-forms-configurator/repositories/DataSetRepository.ts
```diff
@@ -1,7 +1,9 @@
 import { Id } from "../../common/entities/Base";
-import { DataSet } from "../entities/DataSet";
+import { DataElementRef, DataSet, DataSetDetail } from "../entities/DataSet";
 
 export interface DataSetRepository {
     get(): Promise<DataSet[]>;
     getById(id: Id): Promise<DataSet>;
+    getWithElements(id: Id): Promise<DataSetDetail>;
+    createDefaultSection(dataSetId: Id, dataElements: DataElementRef[]): Promise<void>;
```
eperedo: we normally use a `save` method to persist information and in the parameters we use the entity object we're working. In this case `DataSet` so we can reuse always this method anytime we need to modify the entity

src/domain/autogenerated-forms-configurator/repositories/DataSetRepository.ts
```diff
@@ -1,7 +1,9 @@
 import { Id } from "../../common/entities/Base";
-import { DataSet } from "../entities/DataSet";
+import { DataElementRef, DataSet, DataSetDetail } from "../entities/DataSet";
 
 export interface DataSetRepository {
     get(): Promise<DataSet[]>;
     getById(id: Id): Promise<DataSet>;
+    getWithElements(id: Id): Promise<DataSetDetail>;
```
eperedo: A `DataSetRepository` should work with a `DataSet` Entity. If we need a different entity we should create other repository. In this case I think adding `sections` and `dataElements` to the `DataSet` entity makes sense because we need that information later to save it in the server.

src/domain/autogenerated-forms-configurator/usecases/GetDataSetWithElementsUseCase.ts
```diff
@@ -0,0 +1,11 @@
+import { Id } from "../../common/entities/Base";
+import { DataSetDetail } from "../entities/DataSet";
+import { DataSetRepository } from "../repositories/DataSetRepository";
+
+export class GetDataSetWithElementsUseCase {
```
eperedo: This should be `SaveDataSetUseCase` so we can eventually persist any information from the dataSet and avoid to create a single usecase for every attribute that we want to modify.

src/data/autogenerated-forms-configurator/Dhis2DataSetRepository.ts
```diff
@@ -44,6 +45,44 @@ export class Dhis2DataSetRepository implements DataSetRepository {
...
+        const section = {
+            id: generateUid(),
+            name: "Default",
+            dataSet: { id: dataSetId },
+            dataElements: dataElements.map(de => ({ id: de.id })),
+            sortOrder: 0,
+        };
+
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        await this.api.metadata.post({ sections: [section] } as any).getData();
```
eperedo: When we need to persist information we try to do a fetch + save so we don't lose any data in the process.

For this I'd recommend generate a deterministic ID: 

```ts
id: getUid(`Default_${dataSetId}`) // we always generate the same id for this section
``` 

Then we get the section by id and then we create/update according the case. Also by doing this we don't need the explicit cast to `any`

src/data/autogenerated-forms-configurator/Dhis2DataSetRepository.ts
```diff
@@ -44,6 +45,44 @@ export class Dhis2DataSetRepository implements DataSetRepository {
         };
     }
 
+    async getWithElements(id: Id): Promise<DataSetDetail> {
```
eperedo: I think we can reuse the `getById` method. Check my comment in the `DataSetRepository.ts` for more details

eperedo (changes requested):

MiquelAdell:
Thanks @eperedo . Addressed everything in the latest commit:

  - DataSet now includes sections and dataElements directly — DataSetDetail is gone. One entity, one repository.
  - getWithElements is gone too; folded into getById, which now returns the full dataset including sections and elements.
  - createDefaultSection(id, elements) replaced by save(dataSet: DataSet). Passing the full entity means we can persist other dataset attributes later without adding more methods
  each time.
  - Section ID is now getUid('Default_' + dataSet.id). The save implementation fetches by that ID first via metadata.get, then posts — so it handles create and update correctly.
  - as any is gone. D2Section is in the d2-api 2.34 schemas, so PartialModel<D2Section> gives proper typing for the metadata post.
  - Use cases renamed: GetDataSetWithElementsUseCase → GetDataSetUseCase, CreateDefaultSectionUseCase → SaveDataSetUseCase.

src/domain/autogenerated-forms-configurator/entities/DataSet.ts
```diff
@@ -1,15 +1,12 @@
 import { Id, CodedRef } from "../../common/entities/Base";
 
+export type DataElementRef = { id: Id; name: string };
```
eperedo: Not sure If i'm missing something, but I think the name is not being used, right? We can safely remove it.

src/domain/autogenerated-forms-configurator/usecases/CreateDefaultSectionUseCase.ts
```diff
@@ -1,11 +1,10 @@
...
 import { DataSetRepository } from "../repositories/DataSetRepository";
 
-export class CreateDefaultSectionUseCase {
+export class SaveDataSetUseCase {
     constructor(private dataSetRepository: DataSetRepository) {}
 
-    execute(dataSetId: Id, dataElements: DataElementRef[]): Promise<void> {
-        return this.dataSetRepository.createDefaultSection(dataSetId, dataElements);
+    execute(dataSet: DataSet): Promise<void> {
+        return this.dataSetRepository.save(dataSet);
```
eperedo, line 9: I forgot to mention, but here we should validate to only call the `save` method if dataSet has no sections.

eperedo (changes requested):
Looks good, just a couple of small observations

src/data/autogenerated-forms-configurator/Dhis2DataSetRepository.ts
```diff
@@ -41,7 +44,29 @@ export class Dhis2DataSetRepository implements DataSetRepository {
...
+        const sectionId = getUid(`Default_${dataSet.id}`);
+
+        const { sections: existingSections } = await this.api.metadata
+            .get({ sections: { fields: { $owner: true }, filter: { id: { eq: sectionId } } } })
+            .getData();
+
+        const existingSection = existingSections[0];
+        const section: PartialModel<D2Section> = {
+            ...(existingSection ?? {}),
+            id: sectionId,
```
eperedo, line 62: We are not generating the code here, this could be useful if later the user wants to modify any settings for this section because in the json config we reference sections by its code.

Just a comment because I think this is outside of the scope of the task.

eperedo (changes requested):

src/data/autogenerated-forms-configurator/Dhis2DataSetRepository.ts
MiquelAdell: Miquel: yes, I agree it is a good idea to add the code here

MiquelAdell:
Addressed all actionable PR #198 review comments — removed name from DataElementRef, added section-count guard in SaveDataSetUseCase, and cleaned up the API field query

eperedo (approved):

PR #199, opened 2026-06-08
[feature]: Keep numeric inputs consistent across browsers (Firefox space bug)

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/869djz6bn

### :memo: Implementation
`<input type="number">` handled "bad input" inconsistently across browsers: typing a space as a thousands separator (e.g. `1 000`) was stripped to `1000` in Chrome but blanked the whole value in Firefox, silently losing data on save.

Switched `NumberWidget` and `PercentageWidget` to `type="text"` + `inputMode="decimal"` and centralised the behaviour in a new `useNumericInputState` hook so both widgets behave identically on every browser.

#869djz6bn

### DIFF
--- a/src/utils/string.ts
+++ b/src/utils/string.ts
@@ -15,3 +15,17 @@ const INVALID_XML_CHARS_RE = /[\x00-\x08\x0B\x0C\x0E-\x1F\x7F]/g;
 export function replaceInvalidXmlChars(text: string): string {
     return text.replace(INVALID_XML_CHARS_RE, " ");
 }
+
+// Strips every character that cannot appear in a number, keeping only digits,
+// decimal points and minus signs.
+const NON_NUMERIC_CHARS_RE = /[^0-9.-]/g;
+
+export function sanitizeNumericInput(input: string): string {
+    return input.replace(NON_NUMERIC_CHARS_RE, "");
+}
+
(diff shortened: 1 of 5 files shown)


MiquelAdell (approved):

eperedo (changes requested):
Looks good, but I've noticed that the input styles are different now:

Current implementation

image

Before

image

eperedo (changes requested):
Sorry to being so picky here, but now we're losing the padding in the data entry (works well in local).

Image

I think this is related to some css from the data entry

Image

Padding from autogenform is being replaced by the one from the data entry

Image

https://dev.eyeseetea.com/who-dev-41/dhis-web-commons/css/light_blue/light_blue.css?_rev=c19ad76

It should look like this:

Image

Not ideal, but maybe adding important to the padding could fix the problem. Also can we use a min-height of 40 instead 48. I saw this was the previous height in the development branch.

deeonwuli:
> Not ideal, but maybe adding important to the padding could fix the problem. Also can we use a min-height of 40 instead 48. I saw this was the previous height in the development branch.

Thanks @eperedo. This was the fix:

Screenshot 2026-06-17 at 09 27 41

PR #201, opened 2026-07-08
[feature]: Migrate from CRA to Vite and clear Snyk dependency vulnerabilities

deeonwuli:
### :pushpin: References

-   **Issue:** Closes https://app.clickup.com/t/4528615/869drh3vr

### :memo: Implementation
This branch does two things: migrates the build from Create React App to Vite, and clears the outstanding Snyk dependency vulnerabilities on top of it.

**1. CRA → Vite migration** (`feat: migrate from CRA to vite`)
- Replaces react-scripts with Vite (`@vitejs/plugin-react`) and Vitest for tests.
- Single-file form pipeline preserved via `vite-plugin-singlefile` (the data-entry form still ships as one inlined `dist/custom-data-form.html`).
- `vite.config.ts` handles both app variants, the DHIS2 dev proxy, and Node polyfills; CLI scripts updated for the new build.

**2. Dependency security hardening** (Snyk)
Taken from **186 unique issues / 716 vulnerable paths** to **6 unique / 9 paths**, both of which have no upstream fix and neither of which ships in the deployed form.

| Scan | Before | After |
|---|---|---|
| Production (`--prod`, what ships) | criticals + highs present |  9 paths — non-shipping, unfixable |
| Default `snyk test` | 186 unique / 716 paths | 9 paths (3 high, 3 medium)* |

How it was done:
- **Removed dead dependencies:** `react-router-dom`, `postcss-rtl`, `@dhis2/d2-ui-forms`, `ts-purify` (zero source imports; `d2-ui-forms` also dragged an entire vulnerable babel-6 / jsdoc subtree).
- **Upgraded direct deps** to Snyk's targets: `axios`, `lodash`, `@dhis2/app-runtime`, `@eyeseetea/d2-api`, `@eyeseetea/d2-ui-components`, `@material-ui/*`.
- **Added `resolutions`** to force safe transitive versions (parents pin exact versions, so direct bumps alone couldn't dedupe them): `axios 1.18.1`, `lodash 4.18.1`, `@babel/runtime(+corejs3) 7.29.7`, `handlebars 4.7.9`, `@babel/traverse ^7.23.2`, `i18next 19.8.5`, `linkify-it 5.0.1`, `minimatch 3.1.3`, `qs 6.15.2`, `node-fetch 2.6.7`, `glob-parent 5.1.2`, `loader-utils 1.4.2`, `brace-expansion 1.1.13`, `uglify-js 3.14.3`, `minimist 1.2.6`.
- **Kept `styled-jsx` at 4.0.1** — 5.x crashes the browser (`__dirname is not defined`) because `@dhis2/ui` imports `styled-jsx/style`; forced `loader-utils` instead to still clear its vulns.
- **Moved build-only tools to `devDependencies`** (`@dhis2/d2-i18n-extract`, `@dhis2/d2-i18n-generate`, `d2-manifest`, `xlsx`) — correct classification; none are imported by app code or present in the shipped bundle.

**Accepted finding — `i18next` (pinned to `10.6.0`)**
An earlier revision pinned `i18next 19.8.5` to clear a Snyk prototype-pollution high, but that **crashes the app** (`i18next.init is not a function`): `@dhis2/d2-i18n` requires `i18next ^10.3` and calls the old static `init()` at import. The advisory is only patched at ≥19.8.5, which is incompatible with d2-i18n, so we reverted to `10.6.0` (highest compatible 10.x). This knowingly retains the i18next findings — 1 high (`SNYK-JS-I18NEXT-1065979`) + 2 medium (`575536`, `585930`). They are **not reachable here**: i18next only renders static translation keys, with no user-controlled interpolation input (the prototype-pollution / buffer-overflow vector). This is the only deliberately-retained finding.

[snyk-report-2026-07-13.html](https://github.com/user-attachments/files/29978996/snyk-report-2026-07-13.html)

#869drh3vr


MatiasArriola (changes requested):
Thanks @deeonwuli! nice to see this kind of changes!

I've tried running it with `yarn start` but I'm getting an error: `Uncaught TypeError: i18next.init is not a function`. I verified and it is because the resolution pinning `"i18next": "19.8.5",`, after removing that from package.json I got the app to start. 
It happens because `@dhis2/d2-i18n` expects an older version. I didn't review other resolutions but the app seems to load fine after that.

Other comments: 

- Running a global search of `REACT_APP_` still return some results: we want to address these and move to `VITE_`
- `.eslintrc`: do we want to drop `eslint-config-react-app` and disable the `no-undef` rule? I think it is worth reviewing so we are not relaxing too much the lint checks.
- `src/setupProxy.js`: Looks like a leftover file from CRA, we can remove it
- `public/index.html`: same, let's verify it is not used and remove it
- `.nvmrc`: Pinning Node 20 improves things over 16, but Node 20 is already EOL. Now that we are at it, maybe we can use node 22 or even 24, which are the current LTS versions as of now. [reference](https://nodejs.org/en/about/previous-releases)
- `src/data/ExportConstantJsonRepository.ts`: good refactor, just a small naming suggestion: something like `ExportConstantFsRepository` may be a better name than `ExportConstantJsonRepository`, since the important implementation detail here is that the target is the filesystem rather than the json formatting.

Side comments for follow-ups PRs / future work: 

- We are trying to migrate from yarn 1 to yarn 4 in all apps. I'm mentioning it because it is related. Let's keep this PR focused and consider doing so in the future (it is more a PM decision). 
- For the package.json resolutions, I was trying to document the resolutions. [Example](https://github.com/EyeSeeTea/metadata-synchronization/blob/development/RESOLUTIONS.md). This is not yet a practice we stablished so we might be fine without documenting it, but worth considering for exploring in the future and maybe document retroactively. For this I was testing some AI skills that are in the repo but not yet ready.

Also I didn't generate the html and uploaded the custom form to DHIS2. Worth testing since the html generation changed.

Thanks!

deeonwuli:
> I've tried running it with `yarn start` but I'm getting an error: `Uncaught TypeError: i18next.init is not a function`. I verified and it is because the resolution pinning `"i18next": "19.8.5",`, after removing that from package.json I got the app to start. It happens because `@dhis2/d2-i18n` expects an older version. I didn't review other resolutions but the app seems to load fine after that.
> 
> Other comments:
> 
> * Running a global search of `REACT_APP_` still return some results: we want to address these and move to `VITE_`
> * `.eslintrc`: do we want to drop `eslint-config-react-app` and disable the `no-undef` rule? I think it is worth reviewing so we are not relaxing too much the lint checks.
> * `src/setupProxy.js`: Looks like a leftover file from CRA, we can remove it
> * `public/index.html`: same, let's verify it is not used and remove it
> * `.nvmrc`: Pinning Node 20 improves things over 16, but Node 20 is already EOL. Now that we are at it, maybe we can use node 22 or even 24, which are the current LTS versions as of now. [reference](https://nodejs.org/en/about/previous-releases)
> * `src/data/ExportConstantJsonRepository.ts`: good refactor, just a small naming suggestion: something like `ExportConstantFsRepository` may be a better name than `ExportConstantJsonRepository`, since the important implementation detail here is that the target is the filesystem rather than the json formatting.


Thanks for the thorough review @MatiasArriola! Addressed the changes:
- i18next crash: FIxed by reverting to 10.6.0. The Snyk high has no fix compatible with d2-i18n, so I've documented it as an accepted finding in the PR description rather than reintroducing the crash.
- REACT_APP_ → VITE_: zero references left in tracked files.
- setupProxy.js / public/index.html: removed. Also cleared a few more CRA leftovers while there: `craco.config.js`, `babel.config.js`, `.env.production`, and the stale react-app* `.gitignore` entries.
- .nvmrc: bumped to 22 LTS (Node 20 is EOL).
- ExportConstantJsonRepository: renamed to ExportConstantFsRepository, and did the two sibling repos (ImportConstant, ExportDataElementConfig) the same way for consistency.
- `.eslintrc` findings: TypeScript already catches "used something that doesn't exist," and does it better; ESLint's no-undef is a duplicate that false-alarms on TS-only names like interface/type, so the recommended setup is to turn it off and let tsc own it. 

Also regenerated the HTML and uploaded the custom form to DHIS2, it renders fine. Left the yarn 4 migration and `RESOLUTIONS.md` doc as future follow-ups per your note.

MatiasArriola (approved):
Thanks @deeonwuli!! 

I reviewed the eslint changes again and they make sense. Dropping "eslint-config-react-app" is fine since it is a create-react-app related package, and the no-undef case overlaps with tsx in most of the cases.

MiquelAdell (approved):

PR #204, opened 2026-07-23
[FIX] v43 Update data element assigned to more than one data set for the same org unit

anagperal:
### :pushpin: References

-   **Issue:** Closes #869djm0wn [v43 support](https://app.clickup.com/t/4528615/869djm0wn) & #869ee887r [data entry](https://app.clickup.com/t/4528615/869ee887r)

#### Issue

On DHIS2 v43, saving a data value through the single-value endpoint (`POST /api/dataValues`) fails with:

```json
{
    "httpStatus": "Conflict",
    "httpStatusCode": 409,
    "status": "ERROR",
    "message": "Data set detection failed, found multiple sets: `[<datasetA>, <datasetB>]`",
    "errorCode": "E8002"
}
```

This happens whenever a data element rendered by the autogenerated form is assigned to more than one data set for the same org unit — a common, valid metadata setup (e.g. an indicator reused across programs). The app's save request omitted the optional `ds` (data set) parameter, relying on DHIS2 to auto-detect the data set from `de` + `ou`. When the data element is ambiguous, DHIS2 can no longer resolve it and rejects the save.

#### Why v43 and not v41

DHIS2 v43 shipped **[DHIS2-19679](https://dhis2.atlassian.net/browse/DHIS2-19679) ("perf: SQL based data value import")**, a rewrite of the data value import pipeline from per-value Hibernate/JPA saves to a SQL-based bulk path. That path needs to group incoming values by data set (internally via a `DataEntryGroup`) *before* writing, so the data set must be resolved unambiguously up front — hence the new, stricter `E8002` check. Versions before v43 saved each value independently and tolerated the ambiguity without requiring `ds`.

##### Solution

Always send the `ds` (data set) parameter explicitly on data value save requests, removing the app's dependency on DHIS2's implicit data set auto-detection. `dataSetId` is already known by the app at save time (it comes from the data-entry context/URL), so it is threaded through the existing save path instead of relying on the API to infer it.

### :art: Screenshots

v41 OK:

v41

Error v43:

v43_error_file2
v43_error_file
v43_error2
v43_error

Fixed v43:

v43_file_solved
v43_solved

1. PORT=8081 REACT_APP_DHIS2_BASE_URL="" REACT_APP_DHIS2_AUTH="user:password" yarn start
2. http://localhost:8081?orgUnit=ID&dataSet=ID&period=PERIOD where the DS must have data elements shared between different datasets (assigned to more than one data set for the same org unit)
3. Change value in data element


xurxodev (changes requested):
thanks @anagperal

This PR fixes a real, well-diagnosed v43 regression

I verified the fix end-to-end against a v43 instance (`play.im.dhis2.org/stable-2-43-1`) with a data element shared across three datasets. Before the change, `POST /api/dataValues` returned 409 `E8002`, the payload went out as `ou&pe&de&co&value`, and the value was not persisted. After it, the same request returns 201 Created, the payload carries `ds`, and the value is stored.

## 1. Should fix

**`src/data/common/Dhis2DataValueRepository.ts:383` — `delete()` still fails with the same E8002 error.**

`delete()` calls `api.dataValues.postSet({ importStrategy: "DELETE" }, ...)` without a dataset, hitting the same dataset-detection path this PR fixes elsewhere. Confirmed at the API level on v43: `POST /api/dataValueSets?importStrategy=DELETE` without `dataSet` returns `status=ERROR` with `E8002 "found multiple sets: [ce7DSxx5H2I, pBOMPrpg1QX, YFTk3VdO9av]"`, while the same payload plus `"dataSet": "pBOMPrpg1QX"` returns `status=OK`.

The path that reaches it is **deleteRules** (`dataElements.<code>.rules.delete` in the dataStore): `DataEntryItem.tsx:43` → `useDeleteDataValues.ts:38` → `AutogeneratedForm.tsx:321` → `DeleteDataFormValuesUseCase` → `Dhis2DataValueRepository.delete()`. You type into data element A whose rule clears the values of other data elements B; if any B belongs to several datasets, that deletion fails.

The fix mirrors what you already did for `save`: add `dataSetId: Id` to `DataValueRepository.delete()` and `DeleteDataFormValuesUseCase.execute()`, and include `dataSet` in the payload. The call site at `AutogeneratedForm.tsx:321` already has `dataSetId` in scope. Note this endpoint uses `dataSet` in the JSON body, not `ds` as in the query-param form.

One honest caveat on the evidence: I confirmed this at the API level and by following the code path, but **did not reproduce it from the UI**, because the read bug described below blocks it.

**`src/data/common/Dhis2DataValueRepository.ts:69-70` — on v43 the form renders empty even when data exists.**

This one is pre-existing (identical in `development`, not introduced by this PR), but I hit it while testing this change and it is more visible than the E8002 itself, so I'm raising it here rather than as a follow-up.

DHIS2 v43 hoists response fields that have a single common value to the root, instead of repeating them on every `dataValue`. Requesting a specific dataSet/orgUnit/period returns:

```json
{
  "dataSet": "pBOMPrpg1QX", "period": "202601", "orgUnit": "Rp268JB6Ne4",
  "dataValues": [
    { "dataElement": "r6nrJANOqMw", "categoryOptionCombo": "HllvX50cXC0", "value": "42" }
  ]
}
```

The items no longer carry `period` or `orgUnit`, but the selector reads them per item (`dv.orgUnit`, `dv.period`), so both come out `undefined`. Inspecting the React store at runtime, the keys are truncated — `_.compact()` in `DataValueStore.getKey()` drops the empty parts:

```
"r6nrJANOqMw.HllvX50cXC0"                        ← actual
"r6nrJANOqMw.202601.HllvX50cXC0.Rp268JB6Ne4"     ← expected by getKey()
```

`DataValueStore.get()` (`DataValue.ts:143`) always builds the full key, so it never matches and **every field renders empty**. I confirmed the mechanism by requesting two periods at once: `period` then disappears from the root and does appear inside each dataValue — DHIS2 only repeats a field when it varies.

Knock-on effects: `SaveDataFormValueUseCase`'s dedup compares against an empty store, and deleteRules never fire because they read from `dataFormInfo.data.values.store`. The fix would be to fall back to the root-level fields when absent on the item (`dv.period ?? response.period`, same for `orgUnit`), including at `Dhis2DataValueRepository.ts:316`. Note TypeScript doesn't catch this because d2-api's type declares `period` as required.

**PR description — "Notes to the tester" uses the pre-Vite env prefix.**

Step 1 says `REACT_APP_DHIS2_BASE_URL` / `REACT_APP_DHIS2_AUTH`. Since the Vite migration (#201) those are ignored — only `VITE_*` is read (`vite.config.ts:33`, `envPrefix: "VITE_"`), and `README.md:18` already documents the new form. Also, an empty base URL means the `/dhis2` proxy is never registered (`vite.config.ts:57`), so the command as written cannot work. Suggested:

```sh
PORT=8081 VITE_DHIS2_BASE_URL="https://your-instance" VITE_DHIS2_AUTH="user:password" yarn start
```

Worth fixing so the next tester doesn't lose time on it — I did.

## 2. Recommendations non blocking

**`applyToAll()` (`Dhis2DataValueRepository.ts:408`) is structurally exposed to the same failure.** It builds `stDataPost` without a dataset and posts via `postSetAsync`. I could not confirm it: on the demo instance the async import never persists anything even *with* `dataSet` (scheduler appears disabled), so the result was inconclusive rather than negative. Worth checking on a v43 with jobs enabled; by symmetry with `delete()` I'd expect it to need the same treatment.

**No test asserts that `ds` reaches the request.** The spec changes correctly verify `dataSetId` is passed through the use case, but they mock the repository — the actual regression (a missing key in the HTTP payload) would not be caught. A small repository-level test against a mocked API asserting the posted params contain `ds` would lock in the contract that motivated this PR. `DeleteDataFormValuesUseCase.spec.ts` already exists as a precedent for testing at that level.

**`SaveGridWithTotalsValueUseCase.execute` is now 6 positional params, including the unused `_cocId`.** `_cocId` predates this PR (introduced in `15fb577`), but this change shifts it to position 5 and adds a sixth `Id` behind it — transposing two of the string-typed arguments at the call site (`AutogeneratedForm.tsx:357-363`) would compile silently. Sibling use cases in the same directory (`GetDataFormUseCase`, `GetDataFormValuesUseCase`, `GetOrgUnitsUseCase`) switch to an options object at this arity. Under the Boy Scout Rule this would be a reasonable cleanup, but it's your call whether it belongs in a focused bugfix.

**Consider centralizing the dataset key.** Across the five write paths the dataset is spelled by hand each time, and DHIS2 uses two different casings (`ds` for query params, `dataSet` for JSON bodies). A small private helper building the common params would make it harder for a future write path to forget it.

xurxodev (approved):
Thanks @anagperal

